PR #48, opened 2015-04-11
Test to see if installing latest openmpi version helps travis

jakebolewski:


jakebolewski:
I was opening a PR to trigger Travis to run against it, I don't think this actually works yet.

PR #74, opened 2015-06-22
Updated toolchain files for BG/Q, Mira.

justusc:
- Corrected paths and variables for BGQ-Mira-clang-essl.cmake toolchain.
- Added BGQ-Mira-gnu-essl.cmake toolchain file.
- Set the platform name in the above toolchain files to BlueGeneP-static, which also works for BG/Q.

### DIFF
--- a/cmake/toolchains/BGQ-Mira-clang-essl.cmake
+++ b/cmake/toolchains/BGQ-Mira-clang-essl.cmake
@@ -1,9 +1,11 @@
-#set(CMAKE_SYSTEM_NAME BlueGeneQ-static)
+# Set the system name so CMake uses the appropriate platform settings.
+# NOTE: The platforms settings for BlueGeneP are the same for BlueGeneQ 
+set(CMAKE_SYSTEM_NAME BlueGeneP-static)
 
-set(GCC_ROOT  "/bgsys/drivers/ppcfloor/gnu-linux")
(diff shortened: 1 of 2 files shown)


jeffhammond:
Looks good to me other than BlueGeneP in the system name, which I assume is my error from whenever these files were created.

jeffhammond:
I assume we do not have Travis running on BGQ, so waiting for the CI to finish is not useful.

poulson:
Thanks Justus!

justusc:
@jeffhammond Actually, as indicated in the comment, the BlueGeneP system name works just fine for the BlueGene/Q. When I was originally setting up the toolchain files for my project, I had to dig into the platform files since there was no BlueGeneQ platform support. My intention was to extracting the relevant setting and include them in my toolchain files, but I found that their was no need to make any changes. So I simply used the old platform files.

There are BlueGeneQ platform files in CMAKE as of March 2015, but I am not sure if they have even been released yet. It is definitely not available on Mira right now.

jeffhammond:
Yeah, I do not think it is of any consequence to have BlueGeneP there.  It is merely my OCD that objects to it :-)

Given how much trouble I've had with CMake over the years on supercomputers, I only trust toolchain files, no matter what sort of automatic support is claimed in the CMake documentation.

PR #80, opened 2015-07-08
Added configure files for CMake find_package routine

YingzhouLi:


poulson:
Is your text editor automatically deleting trailing spaces? There is a lot of noise in finding what appears to be a single change to ``CMakeLists.txt``. Also, it looks like the boiler-plate ``The interesting stuff goes here`` was left in.

YingzhouLi:
Does this look better? I added back the spaces and removed the comments from templates.

poulson:
Thanks Ryan!

poulson:
Hi @YingzhouLi. I had to make a few changes over the past few days to get this working on recent versions of CMake (for example, PMRRR had to be added to ElTargets from its CMakeLists file). I also moved the configure files into ``cmake/configure_files/``.

Is there a simple example that you've run to test this? If so, I would like to add it to the documentation for the build system.

PR #96, opened 2015-09-15
Fixed a link error and added GridCompare routine for the comparison of two grids (having different memory addresses)

YingzhouLi:


include/El/core/Grid.hpp
```diff
@@ -132,6 +134,92 @@ inline void AssertSameGrids( const Grid& g1, const Grid& g2, Args&... args )
     AssertSameGrids( g2, args... );
 }
 
+inline bool GridCompare
```
poulson: Is there a case where ``GridCompare`` is needed (rather than reusing the same ``Grid`` instance)? This routine seems fairly expensive (in addition to requiring another ``friend`` declaration). If it's to be kept, I think that checking that the relevant communicators are identical rather than just congruent could be faster and less code.

YingzhouLi: I created grid level by level in my hierarchical structure.
The grid created in the constructer does not work for Matrix Combination. However, the local grid based on the same Comm and Group works. That's the reason for me to added this routine to check if these two grids are exactly the same or not.
The current GridCompare seems not returning the correct answer for the following piece of code:
Grid g1(comm, group, height);
Grid g2(comm, group, height);
GridCompare(g1,g2); // this return false.

poulson: In that case, if it isn't possible to simply reuse ``g1``, then I would recommend having ``GridCompare`` check if the output of ``MPI_Comm_compare`` and ``MPI_Group_compare`` are ``MPI_IDENT`` in addition to the height being the same. If the inputs to the constructor are the same, the outputs should be too (and this would eliminate the need for the ``friend`` declaration).

YingzhouLi:
Jack,
I will close this pull request, and please added the link for `AssertNotLocked()`

poulson:
Why close the pull request? I am simply suggesting making ``GridCompare`` much shorter (and faster) via comparing the viewing communicator, owning group, and grid height.

poulson:
FYI: https://github.com/elemental/Elemental/commit/11c298bce1e5ec33ba63b48d805e5692e57411bf

PR #99, opened 2015-10-16
Remove all debug stuff from Travis build

andreasnoack:


poulson:
It didn't go through :8ball:

andreasnoack:
No that was why I asked about `ReadProxy.cpp`. I don't understand why `Elemental.jl` goes through but not this one. I don't see a big difference in the build setup.

poulson:
Are both on the legacy infrastructure? Because the new infrastructure has more RAM.

jakebolewski:
Pinging @tkelman as he is probably the one most familiar with the Travis infrastructure.  

The Elemental travis build keeps running out of memory but the Elemental.jl package (which clones and builds Elemental) does not.

tkelman:
Try `sudo: false` for docker based 12.04 with a bit more memory, or
```
sudo: required
dist: trusty
```
for 14.04 with even more memory

andreasnoack:
Okay. Now we try with "even more memory" so hopefully the build will make it though the proxy curse of dimensionality.

tkelman:
You'll need to adjust the apt-get packages for 14.04.

andreasnoack:
```
[ 38%] Building CXX object CMakeFiles/El.dir/src/core/ReadProxy.cpp.o
[ 39%] Building CXX object CMakeFiles/El.dir/src/core/Grid-C.cpp.o
```
I think this change was sufficient. Thanks @tkelman. There is some clang trusty stuff going wrong, but gcc will probably make it through.

poulson:
@tkelman :100:

jakebolewski:
:racehorse: :gun:

poulson:
@jakebolewski Is the horse Travis or Elemental? :-(

tkelman:
Looks like the failure was just a build time limit. Can you use binaries of more dependencies?

poulson:
@tkelman If so, then it wouldn't really be a fair build test, as Elemental is often responsible for building ScaLAPACK.

poulson:
My honest opinion is that the limitations on Travis make it *almost*, but not quite, suitable for use with Elemental.

tkelman:
> Elemental is often responsible for building ScaLAPACK

Sure, but looks like more of the time is spent on openblas?

andreasnoack:
@tkelman Is there only one core available on the `trusty` machines?

tkelman:
should be 2 I think, there's a table somewhere (maybe travis ci's blog)

poulson:
@tkelman Good point; I forgot about that. This typically only happens on Linux boxes.

PR #103, opened 2015-11-02
This pull request focus on the GridCompare routine

YingzhouLi:
GridCompare function compare the order, height, viewingComm and owningGroup only.


poulson:
Hi Ryan,

The Travis tests seem to be failing due to the curl command being outdated. Would you mind pulling the recent change to `.travis.yml` (submitted by @andreasnoack) which should hopefully fix the issue?

Jack

YingzhouLi:
Hi Jack,

The current pull request passed the Travis CI build.

Yingzhou

include/El/core/environment/decl.hpp
```diff
@@ -265,6 +265,24 @@ DEBUG_ONLY(
     typedef CallStackEntry CSE;
 )
 
+#ifndef EL_RELEASE
+    void LogFileOpen( char* filename );
+
+    std::ostream & LogFileOS();
+
+    void LogFileCoutStr( std::string str );
```
poulson: Would you be okay with shortening `LogFileCoutStr` to just the verb `Log`?

include/El/core/environment/decl.hpp
```diff
@@ -265,6 +265,24 @@ DEBUG_ONLY(
     typedef CallStackEntry CSE;
 )
 
+#ifndef EL_RELEASE
+    void LogFileOpen( char* filename );
```
poulson: Would you be okay with just calling this `OpenLog`?

include/El/core/environment/decl.hpp
```diff
@@ -265,6 +265,24 @@ DEBUG_ONLY(
     typedef CallStackEntry CSE;
 )
 
+#ifndef EL_RELEASE
```
poulson: Would you be okay with using the `DEBUG_ONLY` macro instead?

include/El/core/environment/decl.hpp
```diff
@@ -265,6 +265,24 @@ DEBUG_ONLY(
...
+
+    template<typename... Args>
+    inline void LogFileCout( Args... args )
+    {
+        std::ostringstream str;
+        BuildStream( str, args... );
+        LogFileCoutStr( str.str() );
+    }
+
+    void LogFileClose();
```
poulson: Similarly, could we name this `CloseLog()`?

src/core/global.cpp
```diff
@@ -370,6 +375,12 @@ void Initialize( int& argc, char**& argv )
     const long seed = (secs<<16) | (rank & 0xFFFF);
     ::generator.seed( seed );
     srand( seed );
+
+#ifndef EL_RELEASE
+        char sbuf[50];
+        sprintf(sbuf,"El-Proc%03d.log",rank);
+        LogFileOpen(sbuf);
+#endif
```
poulson: Would you be okay with replacing this segment with:
```
    DEBUG_ONLY(
      ostringstream fileOS;
      fileOS << "El-Proc" << rank << ".log";
      OpenLog( fileOS.str().c_str() );
    )

File: src/core/global.cpp [hunk 6]
```diff
@@ -418,6 +429,10 @@ void Finalize()
         while( ! ::blocksizeStack.empty() )
             ::blocksizeStack.pop();
     }
+
+#ifndef EL_RELEASE
+    LogFileClose();
+#endif
```
poulson: Similarly, for the sake of consistency, could we change this to:
```
DEBUG_ONLY(CloseLog())
```

src/core/global.cpp
```diff
@@ -596,6 +611,24 @@ DEBUG_ONLY(
...
+    }
+
+    std::ostream & LogFileOS() { return ::logFile; }
+
+    void LogFileCoutStr( std::string str )
+    { ::logFile << str << std::endl; }
+
+    void LogFileClose() { ::logFile.close(); }
+#endif
+// LogFile DEBUG_ONLY
```
poulson: To be consistent with the above suggestions, this segment would become:
```
DEBUG_ONLY(
  void OpenLog( const char* filename )
  {
      if( ::logFile.is_open() )
          CloseLog();
      ::logFile.open( filename );
  }

  std::ostream& LogFileOS() { return ::logFile; }

  void Log( std::string str ) { ::logFile << str << std::endl; }

  void CloseLog() { ::logFile.close(); }
)
```

include/El/core/environment/decl.hpp
poulson: Also, I would suggest using `const char* filename` since the string does not need to be changed.

src/core/global.cpp
YingzhouLi: Here I prefer to add leading '0' for the index of process. The alternative way will be
```c++
#include< iomanip>

DEBUG_ONLY(
    ostringstream fileOS;
    fileOS << "El-Proc" << std::setfill('0') << std::setw(3) << rank << ".log";
    OpenLog( fileOS.str().c_str() );
)
```

poulson: That sounds good to me!

YingzhouLi:
Hi Jack,

I have renamed the routines related to Log system. 

Yingzhou

include/El/core/environment/decl.hpp
```diff
@@ -265,6 +265,24 @@ DEBUG_ONLY(
     typedef CallStackEntry CSE;
 )
 
+DEBUG_ONLY(
+    void OpenLog( const char* filename );
+
+    std::ostream & LogOS();
+
+    void LogStr( std::string str );
```
poulson: Is there a reason to still call this `LogStr` instead of `Log`? C++ supports overloading, so the `Str` is implied by the input argument.

YingzhouLi:
Hi Jack,

Thanks for remind me of that. I have merged Log and LogStr.

Yingzhou

PR #112, opened 2015-11-20
Fix issue #111

### LINKED ISSUE #111: Compilation issues on GCC and OpenMPI
Our team has experienced two problems during Elemental build on a cluster. We have used gcc in version 5.2 together with OpenMPI in version 1.6.5

1) In our environment, Scalapack requires providing linking flag for Fortran (GFORTRAN_LIB). However, the cmake script had a bug and this variable has been incorrectly passed to Scalapack because of a missing -D in the command line.
Result: Scalapack is still complaining about undefined variable.

2) Include path for MPI consisting of multiple directories has been incorrectly passed to the ParMETIS. The very same issue has been solved for Scalapack, even comments there explained [the problem](https://github.com/elemental/Elemental/blob/master/cmake/external_projects/ElMath/ScaLAPACK.cmake#L253). Similar solution has been implemented for ParMETIS, but the results of a string replacement was simply unused.
Result: compilation fails because necessary headers (such as mpi.h) can't be found.

Pull request #112 fixes both problems.

mcopik:
Small fixes for cmake scripts.


poulson:
Thanks!

PR #128, opened 2016-01-17
Add Primefield

rhl-:
Issuing a preliminary pull request. I am wondering if travis will CI for me if I do this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,7 +417,7 @@ if(EL_HAVE_QT5)
 
   set(EL_C_CPP_FILES "${EL_C_CPP_SOURCE};${EL_MOC_SRC}")
 else()
-  set(EL_C_CPP_FILES "${EL_C_CPP_SOURCE}")
+  set(EL_C_CPP_FILES "${EL_C_CPP_SOURCE}" include/El/lapack_like/factor/smith/smith.hpp)
 endif()
 
 # Handle the header preparation and installation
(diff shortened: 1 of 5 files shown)


poulson:
Hi Ryan,

It seem that there are several syntax errors in the test related to missing `typename` declarations.

rhl-:
yeah, i believe they are fixed locally. I just wanted to see if this would work.

poulson:
Ah, thanks for clarifying. In that case, I'll wait to review the changesets.

rhl-:
@poulson, in your build system, it looks like you use file globs to build your tests. how do you make it recognize that a file has changed without deleting your build directory and starting from scratch?

rhl-:
also, feel free to say anything if you see anything obvious.

poulson:
@rhl- Touching the `CMakeLists.txt` and rerunning `make` refreshes the globbed list of files and picks up the build on the new executables (and does not require rebuilding the library or pre-existing executables).

rhl-:
@poulson ok, I fixed the compiler errors, but, now I see that I need to export symbols for types like this:

    SparseMatrix< PrimeField< 3> > 

can you point out what I need to do to instantiate these sorts of templates?

poulson:
If the number of types is to become unbounded, which, for practical purposes, it seems to be for your case, it only makes sense to move the `SparseMatrix` implementation entirely into the headers.

rhl-:
@poulson wont this have dramatic effects on users? One could do the same thing for the dense matrix classes as well, and then all your matrix types are header only. I would imagine this would slow down compilation times dramatically. Perhaps an alternative is just exporting a small number of primes? Unless you don't mind moving to header only?

poulson:
With a proper usage of explicit instantiations of common types and extern templates the compile times could perhaps even decrease. But this would need to happen at a library-wide level.

rhl-:
I see. Perhaps we should tackle this first then.

PR #129, opened 2016-01-21
Sparse get

rhl-:
adds get and set functions for [distributed] sparse matrices. there are no unit tests or documentation for this. looking into adding a unit test.

### DIFF
--- a/src/core/DistSparseMatrix.cpp
+++ b/src/core/DistSparseMatrix.cpp
@@ -616,6 +616,21 @@ EL_NO_RELEASE_EXCEPT
     return vals_[localInd];
 }
 
+template< typename T>
+T SparseMatrix<T>::Get( Int row, Int col) const EL_NO_RELEASE_EXCEPT
+{
+    Int index = Offset( row, col);
+    if( index < 0 || index >= Int(vals_.size())){ return T(0); }
+    return Value( Offset( row, col)); 
+}
+
+template< typename T>
+void SparseMatrix<T>::Set( Int row, Int col, T val) const EL_NO_RELEASE_EXCEPT
+{
+    QueueUpdate( row, col, val);
+    ProcessQueues();
+}
+
 template<typename T>
 Int* DistSparseMatrix<T>::SourceBuffer() EL_NO_EXCEPT
 { return distGraph_.SourceBuffer(); }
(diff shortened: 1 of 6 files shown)


rhl-:
i've now added a single test for this, in SparseLDL, it doesn't seem to fail anything..

src/core/DistSparseMatrix.cpp
```diff
@@ -616,6 +616,21 @@ EL_NO_RELEASE_EXCEPT
     return vals_[localInd];
 }
 
+template< typename T>
+T DistSparseMatrix<T>::Get( Int row, Int col) const EL_NO_RELEASE_EXCEPT
+{
+    Int index = Offset( row, col);
+    if( index < 0 || index >= Int(vals_.size())){ return T(0); }
```
poulson: The behavior of `Graph::Offset( Int row, Int col )` is to call `std::lower_bound` on the sorted indices associated with row `row`, and so the returned index could easily be both greater than 0 and less than `vals_.size()` even though there is no edge at `(row,col)`. Thus, I would suggest changing the check to `if( Source(index) != row || Target(index) != col ) { return T(0); }`

rhl-:
I've updated the diff with this change.

poulson:
The `Graph` and `DistGraph` classes' `Source` and `Target` member functions are named `Row` and `Col` within `SparseMatrix` and `DistSparseMatrix`, so I think only minor changes to your PR are needed to get the build working.

rhl-:
Woops...

src/core/SparseMatrix.cpp
```diff
@@ -329,6 +329,23 @@ T SparseMatrix<T>::Value( Int index ) const EL_NO_RELEASE_EXCEPT
...
+{
+    Int index = Offset( row, col);
+    if( Row(index) != row || Col(index) != col){
+            return T(0); 
+    }
+    return Value( Offset( row, col)); 
+}
+
+template< typename T>
+void SparseMatrix<T>::Set( Int row, Int col, T val) EL_NO_RELEASE_EXCEPT
```
poulson: The current implementation of `Set` actually performs an `Update` (that is, if entry `(row,col)` was already nonzero, it adds `val` to it rather than assigning `val`).

rhl-: fixed with a call to QueueZero first.

PR #131, opened 2016-01-27
refactor header symbols. seems to break the the macro system.

rhl-:
This doesn't build, not sure how to fix the macro system..


include/El/blas_like/level3.hpp
```diff
@@ -9,621 +9,25 @@
...
-//       and is obviously not a BLAS routine. However, it is a basic linear
-//       algebra routine making use of Elemental's core data structures, and
-//       so this is the natural placement
-template<typename T>
-void Multiply
-( Orientation orientation,
-  T alpha, const SparseMatrix<T>& A, const Matrix<T>& X,
-  T beta,                                  Matrix<T>& Y );
-
-template<typename T>
```
poulson: It appears you have not replaced this prototype anywhere and that this is the problem.

rhl-: fixed, and the other two missing things at the bottom were added.

rhl-: things build locally.

poulson:
According to the failed Travis build log, https://s3.amazonaws.com/archive.travis-ci.org/jobs/105338873/log.txt, the first error is that `AxpyContract` is not defined within `Symv.hpp`

PR #132, opened 2016-02-08
Multiply sparse matrices

rhl-:
Need to add unit tests.


rhl-:
Lets follow this PR up with a DistGraph*DistMatrix -> DistMatrix variant, which will obviate a DistSparse*DistMatrix -> DistMatrix variant.

include/El/core/DistGraph.hpp
```diff
@@ -11,6 +11,206 @@
 #ifndef EL_CORE_DISTGRAPH_HPP
 #define EL_CORE_DISTGRAPH_HPP
 
-#include <El/core/DistGraph/decl.hpp>
```
poulson: Is there a reason for purposefully breaking the convention of splitting the declarations and implementations into `decl.hpp` and `impl.hpp` files?

include/El/blas_like/level1/EntrywiseMap.hpp
```diff
@@ -160,7 +160,6 @@ void EntrywiseMap
     B.remoteVals_.resize( numRemoteEntries );
     for( Int k=0; k<numRemoteEntries; ++k )
         B.remoteVals_[k] = func(A.remoteVals_[k]);
-    B.multMeta = A.multMeta;
```
poulson: Why remove this?

include/El/core/DistSparseMatrix.hpp
```diff
@@ -11,7 +11,181 @@
 #ifndef EL_CORE_DISTSPARSEMATRIX_HPP
 #define EL_CORE_DISTSPARSEMATRIX_HPP
 
-#include <El/core/DistSparseMatrix/decl.hpp>
-#include <El/core/DistSparseMatrix/impl.hpp>
```
poulson: Same question here again.

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
+/*
+   Copyright (c) 2009-2016, Jack Poulson
```
poulson: I would recommend adding your name as the first author on this file since you introduced it.

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
...
+void TestMultiply(Int m, Int n=100)
+{
+    DEBUG_ONLY(CallStackEntry cse("TestMultiply"))
+    SparseMatrix<T> A;
+    Identity( A, m, m);
+    Matrix<T> B( m, n);
+    Bernoulli( B, m, n);
+    auto G = A.Graph();
+    Matrix<T> C(m,n),D(m,n);
+    Multiply(NORMAL, T(1.0), A, B, T(0.0), C);
```
poulson: I would recommend just `T(1)` and `T(0)`.

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
...
+    Multiply(NORMAL, T(1.0), G, B, T(0.0), D);
+    Axpy(-1.0, C, D);
+    auto nrm = FrobeniusNorm(D);
+    std::cout << "error = " << nrm << std::endl;
+    if( nrm > 1e-16){ throw std::runtime_error("Sparse(I)*x != Graph(I)*x"); }
```
poulson: It would be better to poll the machine epsilon via `limits::Epsilon<Real>()` and incorporate the appropriate norms. I would recommend looking in Higham's Accuracy and Stability of Numerical Algorithms for rigorous bounds for this.

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+int main( int argc, char* argv[] )
+{
+    Environment env( argc, argv );
+    mpi::Comm comm = mpi::COMM_WORLD;
+    try
+    {
+	Int m = 1;
+	for(int e = 1; e < 4; ++e){
```
poulson: Consistent for loop style, please? The rest of the library would use:
```
for( Int e=1; e<4; ++e )
{
```

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
...
+int main( int argc, char* argv[] )
+{
+    Environment env( argc, argv );
+    mpi::Comm comm = mpi::COMM_WORLD;
+    try
+    {
+	Int m = 1;
+	for(int e = 1; e < 4; ++e){	
+		m *= 10;	
+		std::cout << " m = " << m << std::endl;
```
poulson: There is a routine named `Output` that Elemental introduced that is somewhat preferred to `std::cout`.

tests/blas_like/Multiply.cpp
poulson: Also, calling `RuntimeError` is preferred to `throw std::runtime_error` since the former allows one to easily compose nontrivial messages, such as `RuntimeError("The matrix height was ",C.Height())`.

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
...
+		RunTests( m);
```
poulson: Also, what about distributed tests?

include/El/core/DistGraph.hpp
rhl-: It's been a very long time since I initially did these changes. However, this convention is a bit odd. It is not as if the impl.hpp is a .cpp file. 

I propose that we (as much as possible) move to single header-only implementations.  

Simply moving the decl/impl into one header is asking the pre-processor to do less work.

include/El/blas_like/level1/EntrywiseMap.hpp
rhl-: I think because a few lines early we issue:

`  B.distGraph_ = A.distGraph_;`

which assigns the multMeta objects.

include/El/core/DistGraph.hpp
poulson: In practice there is a lot of movement between the `impl.hpp` and an associated source file. For `DistMatrix`, there have been at least five such movements. If nothing else, consistency is important.

include/El/blas_like/level1/EntrywiseMap.hpp
poulson: Ah, agreed.

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
+/*
```
rhl-: I am only covering the routines I added for now. I have not yet added any Distributed multiply's. WHen I add then they will get tests, and i'll try at that point to cover more routines.

include/El/core/DistGraph.hpp
jedbrown: @rhl- In general, you're asking to spend way more time compiling if you combine them. It is very common for header dependencies and transitive dependencies (header or source) to need only the declarations, which means that all those compilation units don't need to pull in the implementation and don't need to be recompiled any time the impl.hpp changes (usually much more common than decl.hpp changes). We saw a tangible improvement in PETSc compile times (especially incremental) when we split some headers.

rhl-: Hi Jed,

Not sure this is accurate. This is a pure header only implementation, with
separate sources, so you have to spend more time because its not all in one
place. However, I agree that there is value for only having signatures. If
impl.hpp was impl.cpp, then what you are saying is accurate.

Right now the headers extern template instantiations for a zillion things,
this is assuredly makes the builds *very* slow, because template
_instantiation_ is relatively slow.

I'm trying to back out these changes for now.

Best,
-rhl

jedbrown: Can you quantify that extra time to pull in both parts of the file?

Also, I really am referring to splitting headers into the part that only declares the types and the part that contains the API associated with that type. It is common for dependent headers to only need the declaration -- only compilation units that actually manipulate that type need to include the rest of the header. You'll find many such examples in `/usr/include/**/*{fwd,decl,type,types}.h{,pp}`. Boost is insistent about factoring headers in that way, for example.

include/El/core.hpp
```diff
@@ -235,7 +235,7 @@ class DistMatrix;
 #include <El/core/Matrix/decl.hpp>
 #include <El/core/Graph/decl.hpp>
 #include <El/core/DistMap/decl.hpp>
-#include <El/core/DistGraph/decl.hpp>
+#include <El/core/DistGraph.hpp>
```
poulson: Why include DistGraph.hpp rather than DistGraph/decl.pp?

include/El/core/DistGraph/decl.hpp
```diff
@@ -134,6 +173,10 @@ class DistGraph
     // total number of edges divided by the number of processes
     double Imbalance() const EL_NO_RELEASE_EXCEPT;
 
+    mutable DistGraphMultMeta multMeta;
+    DistGraphMultMeta InitializeMultMeta() const;
+
+
```
poulson: Why the extra line of white space?

include/El/core/DistSparseMatrix/decl.hpp
```diff
@@ -11,44 +11,8 @@
 #ifndef EL_CORE_DISTSPARSEMATRIX_DECL_HPP
 #define EL_CORE_DISTSPARSEMATRIX_DECL_HPP
 
-namespace El {
+namespace El  {
```
poulson: Why the extra single space?

src/blas_like/level3/Multiply.cpp
```diff
@@ -14,6 +14,60 @@
...
+            const Int eStart = rowOffsets[i];
+            const Int eStop = rowOffsets[i+1];
+            for( Int e=eStart; e<eStop; ++e )
+                sum += x[colIndices[e]];         
+            y[i] = alpha*sum + beta*y[i];
+        }
+    }
+    else
+    {
+        const bool conj = ( orientation == ADJOINT );
```
poulson: There is no need to distinguish between `orientation == TRANSPOSE` and `orientation == ADJOINT` if the values are binary (or even if they are arbitrary real). It would be a good idea to delete the `if( conj )` branch.

src/blas_like/level3/Multiply.cpp
```diff
@@ -204,6 +258,78 @@ void MultiplyCSR
...
+                const Int eStop = rowOffsets[i+1];
+                for( Int e=eStart; e<eStop; ++e )
+                    sum += X[colIndices[e]+k*ldX];
+                Y[i+k*ldY] = alpha*sum + beta*Y[i+k*ldY];
+            }
+        }
+    }
+    else
+    {
+        const bool conj = ( orientation == ADJOINT );
```
poulson: The same comment here about conjugation.

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+int main( int argc, char* argv[] )
+{
+    Environment env( argc, argv );
+    mpi::Comm comm = mpi::COMM_WORLD;
+    try
+    {
+	Int m = 1;
+	for( Int e = 1; e<4; ++e ){
```
poulson: Could I convince you to move the curly brace to the start of the next line for the sake of consistency? Pretty, please?

tests/blas_like/Multiply.cpp
```diff
@@ -0,0 +1,53 @@
...
+	Int m = 1;
+	for( Int e = 1; e<4; ++e ){	
+		m *= 10;	
+		Output(" m = ",m,"\n");
+		RunTests( m);
+	}
+    }
+    catch( exception& e ) { ReportException(e); }
+    return 0;
+}
```
poulson: Otherwise, LGTM.

tests/blas_like/Multiply.cpp
rhl-: just this once, but, that is something we need to enforce programmatically, via tooling like clang-tidy.

src/blas_like/level3/Multiply.cpp
rhl-: done

rhl-: weird, its not done.. hmm..

PR #135, opened 2016-02-15
Fix includes

rhl-:
used the tool "include what you use" to make each header include the things that it uses directly, with a bias towards forward declarations.


rhl-:
crud! master has changed since i started doing this dramatically. Looks like it's all good things though. luckily I built this all as scripts, so relatively easy to redo.

PR #137, opened 2016-02-17
Implemented robust multi-shift triangular solve and triangular eigensolver

timmoon10:
See functions SafeMultiShiftTrsm and TriangEig. These functions are currently only implemented for local matrices.


poulson:
Hi Tim,

This is great! What sort of error/timings vs. LAPACK are you seeing for eigenvectors of a Fox-Li matrix of size 1000, etc.?

I am guessing that the unblocked routine could be substantially faster using direct buffer manipulation rather than `Get`/`Set` calls.

Also, I would suggest adding some sort of notice into every file that you modified with your name! (Whether it is through copyright or an "originally written by" statement).

Lastly, this is a bit nit-picky, but the spacing seems to be a bit inconsistent in several places.

timmoon10:
Hi Jack,
[Here](https://github.com/elemental/Elemental/files/145969/TriangEigBenchmark.tar.gz) is a simple program that compares errors and timings for nonsymmetric eigensolvers in Elemental and MKL. I test with Gaussian, Fox-Li, and Grcar matrices, all in complex double. Both Elemental and MKL produce results accurate up to machine precision for all test matrices. Testing a 5000x5000 Fox-Li matrix on nesterov, Elemental took 75.4 s while MKL took 148.1 s. In more detail:

Step | Elemental | MKL
-------|--------------|------
Upper Hessenberg form | (Schur) | 32.087 s
Householder transforms to matrix | (Schur) |  4.692 s
Schur decomposition | 66.089 s | 28.988 s
Triangular eigensolver | 6.324 s | 82.294 s
Back substitution | 2.988 s | (Triangular Eigensolver)

Elemental and MKL take similar amounts of time to compute a Schur factor (~66 s). However, Elemental's triangular eigensolver is about 8.7x faster (9.4 s vs. 82.3 s).

poulson:
Tim, this is absolutely amazing and definitely publication-worthy.

poulson:
And, for what it's worth, the difference in runtime between the Schur decomposition and MKL is easily fixed since the Schur decomposition is currently just a wrapper to ScaLAPACK's earlier version of Hessenberg QR that does not use Aggressive Early Deflation (you could enable Aggressive Early Deflation by modifying the `SchurCtrl` data structure, but I had disabled it by default due to bugs on some architectures).

jiahao:
How soon can we have a Julia version? :)

timmoon10:
I can get it done in a few days.

timmoon10:
Here are some timing results with the new commit (TriangEig uses a shortcutted version of SafeMultiShiftTrsm). All of the trials I conducted produced eigendecompositions that were accurate to machine precision. The experimental setup is the same (5000x5000 Fox-Li matrix running on nesterov).

Step | Elemental | MKL
-------|--------------|--------
Upper Hessenberg reduction | (Schur) | 77.30 s
Householder to full matrix | (Schur) | 6.98 s
Schur factorization | 103.682 s | 26.07 s
Triangular eigensolver | 3.10 s | 188.41 s
Back-transform | 1.63 s | (Triangular eigensolver)

In the end, Elemental takes 108.4 s and MKL takes 259.0 s. The triangular eigensolver sees a 40x speedup relative to MKL's ZTREVC. It is strange that MKL and the Elemental Schur factorization see a 2x slowdown relative to the previous experiment, so I am a little suspicious of these results.

PR #139, opened 2016-03-10
BGQ Juqueen Toolchain file added

SebastianAchilles:
This is the toolchain file that I have used to install Elemental on the Blue gene Q system JUQUEEN @ FZ Juelich.


poulson:
Thanks Sebastian!

PR #140, opened 2016-04-15
LeftLLL Precision Drop

restrin:
This change separates the templates for the QR factorization and basis matrix to allow for the QR factorization to be kept in low precision. Complex LLL cannot be instantiated due to conversion issues between complex datatypes are resolved.

Main changes:
- Support for separate datatypes for the basis matrix and QR factorization
- New flag (minColThresh) in LLLCtrl to allow for BLAS3 for size reduction only when k>=ctrl.minColThresh. Currently default is 0 (to be determined later).
- New flag (unsafeSzReduct) in LLLCtrl to disable checks in size reduction loop when using low precision QR.
- New flag (precisionFudge) in LLLCtrl to control how aggressively to drop precision in recusive LLL.
- Modification to LeftLLL size reduction loop to reduce number of norm computations and safely perform size reduction when QR is in low precision.

These changes result in recusive LLL on SVPChallenge100 to compute in ~10s (QR in double) instead of >1m (QR in BigFloat).


include/El/lattice.h
```diff
@@ -52,6 +52,9 @@ typedef struct
     ElLLLVariant variant;
     bool recursive;
     ElInt cutoff;
+    float precisionFudge;
+    ElInt minColThresh;
+    bool unsafeSzReduct;
```
poulson: This convention isn't written down anywhere, but I would recommend `unsafeSizeReduce` over `unsafeSzReduct`, as it is only one more letter and uses full words

include/El/lattice.hpp
```diff
@@ -126,6 +126,14 @@ struct LLLCtrl
     bool recursive=false;
     Int cutoff=10;
 
+    // Fudge factor for determining whether to drop precision
+    Real precisionFudge=Real(2);
+    
+    Int minColThresh = 0;
+
+    // Ignore precision limits for QR factorization?
+    bool unsafeSzReduct=false;
```
poulson: The same would obviously apply here.

src/lattice-C.cpp
```diff
@@ -21,6 +21,9 @@ ElError ElLLLCtrlDefault_s( ElLLLCtrl_s* ctrl )
     ctrl->variant = EL_LLL_NORMAL;
     ctrl->recursive = false;
     ctrl->cutoff = 10;
+    ctrl->precisionFudge = 2.0f;
+    ctrl->minColThresh = 0;
+    ctrl->unsafeSzReduct = false;
```
poulson: I'm impressed that you added the C struct initialization!

poulson: I think the Python interface on top of the C interface will be broken with this changeset until the equivalent changes are propagated though.

include/El/lattice/LLL/Left.hpp
```diff
@@ -268,14 +292,34 @@ bool Step
...
+            break;
+        }
+
+        colUpdated = false;
+
+        Real newNorm = lll::Norm2<Z,F>(B, k);
+        auto rCol  = QR( ALL, IR(k) );
+        Real rNorm = El::FrobeniusNorm(rCol);
+
+        if (Abs(newNorm - rNorm)/newNorm >= thresh)
```
poulson: This is pedantic, but for the sake of consistency, I would recommend using `if( condition )` rather than `if (condition)` since the former is what the rest of the library uses (and there is a slippery slope).

include/El/lattice/LLL/Left.hpp
```diff
@@ -268,14 +292,34 @@ bool Step
...
+        {
+            if ( ctrl.progress )
```
poulson: same here

poulson:
Add appropriate copyright notices and/or AUTHORS changes and I think this is ready to merge!

poulson:
LGTM!

PR #146, opened 2016-06-08
Fixing args for MKL Trrk & xgemmt routines

rocanale:
Note that the xgemmt routines only work for mkl 2016. Older versions fail when linking the library.


poulson:
The following commit should fix the issue that you raised: https://github.com/elemental/Elemental/commit/37bae0ef27d56ad4c9e98e79f690b8a9c120016c

rocanale:
Perfect, now it builds with the icc 14.0 MKL distribution

PR #158, opened 2016-08-02
Test with Clang. Test on Ubuntu 14.04 to get MPICH 3. Use ccache to speed up tests.

andreasnoack:
It took a while for me to figure out how to set up `ccache` on Travis but it seems to work now. Build times are down to 10-15 minutes. This will probably also make it possible to test with a recent version of OpenMPI which requires that we build it first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,16 @@
+sudo: required
+dist: trusty
+
+notifications:
+  email: false
+
 language: generic
 
 env:
   global:
-    - MPI=mpich2
+    - MPI=mpich
     - F77=gfortran-5
+    - PATH=/usr/lib/ccache/:$PATH
 #  - MPI=openmpi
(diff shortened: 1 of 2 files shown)


rhl-:
Awesome!!

andreasnoack:
Looks like `push` and `pr` are not sharing the build cache.

andreasnoack:
Okay. Looks like the caching is working now on both `pr` and `push`.

andreasnoack:
Hm. Just realized that the master branch is not inheriting the cache from the PR but will need to build its own. New PRs should be able to use the master branch cache and therefore be much faster but it might be a problem to finish the first build after the merge. In the PR, I had to disable the tests while the cache was built and then update the PR to reenable the tests. We might have to do something similar on the master branch, i.e. disable the tests temporarily and then reenable them when the cache has been built.

PR #159, opened 2016-08-03
Temporarily disable tests while building up the cache on Travis.

andreasnoack:
Test should be reenabled as soon as both the GCC and Clang builds have succeeded
on the master branch.


andreasnoack:
It might make sense to merge this right away instead of spending time on building it here on the wrong branch. The important part is that the build succeeds once on the master branch.

PR #162, opened 2016-08-06
Fixed binary install targets

rhl-:
this is necessary to make all the binaries be on the PATH by default, and also make the rpm glob understand the files it is installing without manually specifying each subdirectory.


rhl-:
I think I am quite close to finally finishing the rpm spec: 

https://copr-be.cloud.fedoraproject.org/results/rhl/elemental/fedora-22-x86_64/00438784-elemental/build.log.gz

Im running the build off of this branch right now: 

https://copr.fedorainfracloud.org/coprs/rhl/elemental/build/438827/

poulson:
Is this changeset trying to install all binaries into the same directory? If so, I'm not okay with that being the default as it would be a huge mess. Perhaps we could have a CMake option which chooses between the two which you could activate for building RPM's.

rhl-:
Yeah, I suppose I could fix it do both with an option to simply disable install subdirectories.

rhl-:
Let me see how this goes, since each RPM build takes 1 hr its taking a while to get this right.

CMakeLists.txt
```diff
@@ -516,7 +516,9 @@ if(BUILD_SHARED_LIBS)
       DESTINATION python/El/core)
   endif()
 endif()
-
+if(not NO_BINARY_SUBDIRECTORIES)
```
poulson: The capitalization of 'not' here seems to be inconsistent with the rest of the change set.

CMakeLists.txt
```diff
@@ -606,7 +608,10 @@ if(EL_TESTS)
         set_target_properties(tests-${TYPE}-${TESTNAME} PROPERTIES
           LINK_FLAGS ${EL_LINK_FLAGS})
       endif()
-      install(TARGETS tests-${TYPE}-${TESTNAME} DESTINATION ${CMAKE_INSTALL_BINDIR})
+      if( NOT NO_BINARY_SUBDIRECTORIES)
```
poulson: There is a space after the parenthesis here but not elsewhere.

CMakeLists.txt
```diff
@@ -649,8 +654,11 @@ if(EL_EXAMPLES)
         set_target_properties(examples-${TYPE}-${EXNAME} PROPERTIES
           LINK_FLAGS ${EL_LINK_FLAGS})
       endif()
+      if( NOT NO_BINARY_SUBDIRECTORIES)
```
poulson: same here.

CMakeLists.txt
```diff
@@ -649,8 +654,11 @@ if(EL_EXAMPLES)
...
+      	set(EXAMPLE_DIR examples/${TYPE})
+      endif()
       install(TARGETS examples-${TYPE}-${EXNAME}
-	      DESTINATION ${CMAKE_INSTALL_BINDIR})
+	      DESTINATION ${CMAKE_INSTALL_BINDIR}/${EXAMPLE_DIR})
       add_test(NAME Examples/${TYPE}/${EXNAME} WORKING_DIRECTORY "${OUTPUT_DIR}"
         COMMAND examples-${TYPE}-${EXNAME})
     endforeach()
```
poulson: Other than capitalization and spacing, it LGTM.

PR #167, opened 2016-08-25
Update CMakeLists.txt

rhl-:
When we install we rename files to there unique target names, this prevents later collision.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -618,7 +618,7 @@ if(EL_TESTS)
       if(NOT NO_BINARY_SUBDIRECTORIES)
       	set(TEST_INSTALL_DIR test/${TYPE})
       endif()
-      install(TARGETS tests-${TYPE}-${TESTNAME} DESTINATION ${CMAKE_INSTALL_BINDIR}/${TEST_INSTALL_DIR})
+      install(TARGETS tests-${TYPE}-${TESTNAME} DESTINATION ${CMAKE_INSTALL_BINDIR}/${TEST_INSTALL_DIR} RENAME tests-${TYPE}-${TESTNAME})
       if(NOT TESTNAME STREQUAL "SparseLDLRange") #Skip tests that can time out
         add_test(NAME Tests/${TYPE}/${TESTNAME} 
           WORKING_DIRECTORY "${TEST_DIR}" COMMAND tests-${TYPE}-${TESTNAME})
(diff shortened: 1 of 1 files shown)


poulson:
If I understand correctly, this changes the usual build system to, for example, change the produced binary `./bin/tests/lapack_like/QR` to `./bin/tests/lapack_like/tests-lapack_like-QR`. This seems like a good idea if the directory structure is squashed, but not so much otherwise. Should we add another CMake option for squashing the output directories and branch based on it?

rhl-:
Yes but only on install so only after make install

CMakeLists.txt
```diff
@@ -607,17 +607,19 @@ if(EL_TESTS)
       set_source_files_properties("${DRIVER}" PROPERTIES
         OBJECT_DEPENDS "${PREPARED_HEADERS}")
       target_link_libraries(tests-${TYPE}-${TESTNAME} El)
+      set(TEST_OUTPUT_NAME ${TESTNAME})
+      if(NOT NO_BINARY_SUBDIRECTORIES)
```
poulson: I think this would be easier to read if it was of the form:
```
if(NO_BINARY_SUBDIRECTORIES)
  set(TEST_OUTPUT_NAME ${TESTNAME})
  set(TEST_OUTPUT_DIR "")
else()
  ...
endif()
```
and if we added a short note on the CMake option `NO_BINARY_SUBDIRECTORIES` at the top of the file.

CMakeLists.txt
```diff
@@ -653,8 +655,13 @@ if(EL_EXAMPLES)
       set_source_files_properties("${DRIVER}" PROPERTIES
         OBJECT_DEPENDS "${PREPARED_HEADERS}")
       target_link_libraries(examples-${TYPE}-${EXNAME} El)
+      set(EXAMPLE_OUTPUT_NAME ${EXNAME})
+      if(NOT NO_BINARY_SUBDIRECTORIES)
```
poulson: Same here.

CMakeLists.txt
```diff
@@ -607,17 +607,20 @@ if(EL_TESTS)
       set_source_files_properties("${DRIVER}" PROPERTIES
         OBJECT_DEPENDS "${PREPARED_HEADERS}")
       target_link_libraries(tests-${TYPE}-${TESTNAME} El)
+      if(NOT NO_BINARY_SUBDIRECTORIES)
```
poulson: It is generally considered good form to avoid double-negatives if it doesn't change the control flow too much. For example, if you switched the `if` and `else` branches, it goes away.

CMakeLists.txt
```diff
@@ -653,17 +656,20 @@ if(EL_EXAMPLES)
       set_source_files_properties("${DRIVER}" PROPERTIES
         OBJECT_DEPENDS "${PREPARED_HEADERS}")
       target_link_libraries(examples-${TYPE}-${EXNAME} El)
+      if(NOT NO_BINARY_SUBDIRECTORIES)
```
poulson: Same here. Otherwise, LGTM.

CMakeLists.txt
rhl-: I'll follow this up with a fix for this, because then I also need to update the spec file. In the interest of getting a build, I am going to merge.

poulson:
What I was requesting was a trivial change with no external implications...

rhl-:
I'm not sure what the problem is, the double negation has been there for
some time. There should be no external effects.

poulson:
The point is double negation is bad form. I will fix it when I have computer access again tonight.

rhl-:
I'll take care of it...

Issue #168, opened 2016-08-26
El::MaxAbs(const AbstractDistMatrix<F> &A) not implemented, but method definition is


davidsblom:
When I call the routine:
``` c++
double maxAbs = El::MaxAbs( matrix );
```
the compilation is successfull, but the linker cannot find the implementation.

```
libRBFMeshMotionSolver.so: undefined reference to `El::BaseHelper<double>::type El::MaxAbs<double>(El::AbstractDistMatrix<double> const&)'
collect2: error: ld returned 1 exit status
```

I digged around in the code, and found that it is still a todo. 

I've implemented my own version like this by reusing some code of the Max() method.

``` c++
template<typename Real>
Real MaxAbs( const El::DistMatrix<Real> & A )
{
    Real value = 0;

    if ( A.Participating() )
    {
        // Store the index/value of the local pivot candidate
        const El::Int mLocal = A.LocalHeight();
        const El::Int nLocal = A.LocalWidth();
        const Real * ABuf = A.LockedBuffer();
        const El::Int ALDim = A.LDim();

        for ( El::Int jLoc = 0; jLoc < nLocal; ++jLoc )
            for ( El::Int iLoc = 0; iLoc < mLocal; ++iLoc )
                value = El::Max( std::abs( value ), ABuf[iLoc + jLoc * ALDim] );

        value = El::mpi::AllReduce( value, El::mpi::MAX, A.DistComm() );
    }

    El::mpi::Broadcast( value, A.Root(), A.CrossComm() );
    return value;
}
```

rhl-:
Issue a PR and we can review it :)

davidsblom:
Fixed in #169. Thanks!

PR #169, opened 2016-08-26
Add implementation of MaxAbs routine

davidsblom:
Refer to issue #168

### DIFF
--- a/src/blas_like/level1/MaxAbs.cpp
+++ b/src/blas_like/level1/MaxAbs.cpp
@@ -45,9 +45,9 @@ namespace El {
             const Int ALDim = A.LDim();
             for( Int jLoc=0; jLoc<nLocal; ++jLoc )
                 for( Int iLoc=0; iLoc<mLocal; ++iLoc )
-                    value = Max(value,ABuf[iLoc+jLoc*ALDim]);
+                    value = Max(std::abs(value),ABuf[iLoc+jLoc*ALDim]);
 
-            value = mpi::AllReduce( std::abs(value), mpi::MAX, A.DistComm() );
+            value = mpi::AllReduce( value, mpi::MAX, A.DistComm() );
         }
         mpi::Broadcast( value, A.Root(), A.CrossComm() );
         return value;
--- a/src/blas_like/level1/MaxAbs.cpp
+++ b/src/blas_like/level1/MaxAbs.cpp
@@ -11,46 +11,58 @@
 
 namespace El {
 
-    template<typename Real,typename>
-    Real MaxAbs( const Matrix<Real>& A )
+template<typename Real>
+Base<Real> MaxAbs( const Matrix<Real>& A )
+{
+    DEBUG_CSE
+    const Int m = A.Height();
+    const Int n = A.Width();
+    const Real* ABuf = A.LockedBuffer();
+    const Int ALDim = A.LDim();
+
+    Real value = 0;
+    for( Int j=0; j<n; ++j )
+        for( Int i=0; i<m; ++i )
+            value = Max(abs(value),ABuf[i+j*ALDim]);
+    return value;
+}
+
+template<typename Real>
+Base<Real> MaxAbs( const AbstractDistMatrix<Real>& A )
+{
+    DEBUG_CSE
+    DEBUG_ONLY(
+      if( !A.Grid().InGrid() )
+          LogicError("Viewing processes are not allowed");
+    )
+    Real value = 0;
+    if( A.Participating() )
     {
-        DEBUG_CSE
-        const Int m = A.Height();
-        const Int n = A.Width();
+        // Store the index/value of the local pivot candidate
+        const Int mLocal = A.LocalHeight();
+        const Int nLocal = A.LocalWidth();
         const Real* ABuf = A.LockedBuffer();
         const Int ALDim = A.LDim();
+        for( Int jLoc=0; jLoc<nLocal; ++jLoc )
+            for( Int iLoc=0; iLoc<mLocal; ++iLoc )
+                value = Max(abs(value),ABuf[iLoc+jLoc*ALDim]);
 
-        Real value = 0;
-        for( Int j=0; j<n; ++j )
-            for( Int i=0; i<m; ++i )
-                value = Max(std::abs(value),ABuf[i+j*ALDim]);
-        return value;
+        value = mpi::AllReduce( value, mpi::MAX, A.DistComm() );
     }
-
-    template<typename Real,typename>
-    Real MaxAbs( const AbstractDistMatrix<Real>& A )
-    {
-        DEBUG_CSE
-        DEBUG_ONLY(
-          if( !A.Grid().InGrid() )
-              LogicError("Viewing processes are not allowed");
-        )
-        Real value = 0;
-        if( A.Participating() )
-        {
-            // Store the index/value of the local pivot candidate
-            const Int mLocal = A.LocalHeight();
-            const Int nLocal = A.LocalWidth();
-            const Real* ABuf = A.LockedBuffer();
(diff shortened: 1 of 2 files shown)


rhl-:
Please add a unit test.

LGTM.

davidsblom:
I'll add a unit test. I'll also change the implementation a bit. Noticed that the implementation is slightly different from the definition in `decl.hpp`

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -0,0 +1,56 @@
...
+            for( Int jLoc=0; jLoc<nLocal; ++jLoc )
+                for( Int iLoc=0; iLoc<mLocal; ++iLoc )
+                    value = Max(std::abs(value),ABuf[iLoc+jLoc*ALDim]);
+
+            value = mpi::AllReduce( value, mpi::MAX, A.DistComm() );
+        }
+        mpi::Broadcast( value, A.Root(), A.CrossComm() );
+        return value;
+    }
+
```
poulson: I would recommend following the format of the Max routine: https://github.com/elemental/Elemental/blob/05f8445214f54149332ea12a739b7c10f407ecaa/src/blas_like/level1/Max.cpp#L128

For example, template functions must be explicitly instantiated. Elemental has a style for handling this that defines a `PROTO` macro, sets some definitions to guide which types to instantiate, and then includes the file `El/macros/Instantiate.h` to handle the instantiations given the macro(s) and guidance definitions.

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -0,0 +1,56 @@
...
+   This file is part of Elemental and is under the BSD 2-Clause License,
+   which can be found in the LICENSE file in the root directory, or at
+   http://opensource.org/licenses/BSD-2-Clause
+*/
+#include <El-lite.hpp>
+#include <El/blas_like/level1.hpp>
+
+namespace El {
+
+    template<typename Real,typename>
```
poulson: Elemental's style is to not indent after namespace declarations. Often namespaces can be several deep and it becomes untenable to add four spaces after each. Cf. Google's C++ style guide for this decision.

davidsblom:
Thanks for the comments. I've added some straightforward unit tests. I've used the `assert` method to check if everything is all right. I guess `LogicError` should be used instead?

davidsblom:
Don't merge it yet if you are satisfied. I'm currently adding some more unit tests, and discovered a bug.

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -11,46 +11,58 @@
 
 namespace El {
 
-    template<typename Real,typename>
-    Real MaxAbs( const Matrix<Real>& A )
+template<typename Real>
```
poulson: This routine would work equally well with complex inputs as well (in which case the Elemental convention would be to change `Real` to `F`).

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -11,46 +11,58 @@
...
+    DEBUG_CSE
+    const Int m = A.Height();
+    const Int n = A.Width();
+    const Real* ABuf = A.LockedBuffer();
+    const Int ALDim = A.LDim();
+
+    Real value = 0;
+    for( Int j=0; j<n; ++j )
+        for( Int i=0; i<m; ++i )
+            value = Max(abs(value),ABuf[i+j*ALDim]);
```
poulson: The `std::abs` function only works with standard C++ datatypes, such as `int`, `float`, `std::complex<float>`, and `std::complex<double>`. But they don't work for Elemental's plethora of other datatypes, such as `El::DoubleDouble`, `El::QuadDouble`, `El::Quad, `El::BigFloat`, `El::Complex<El::DoubleDouble>`, `El::Complex<El::QuadDouble>`, `El::Complex<El::Quad>`, or `El::Complex<El::BigFloat>`. For this reason it is necessary to use `El::Abs`. I am assuming that you built Elemental without support for QD, __float128, or MPFR.

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -11,46 +11,58 @@
...
+    return value;
+}
+
+template<typename Real>
```
poulson: Same comment on `s/Real/F/` here.

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -11,46 +11,58 @@
...
-        const Int m = A.Height();
-        const Int n = A.Width();
+        // Store the index/value of the local pivot candidate
+        const Int mLocal = A.LocalHeight();
+        const Int nLocal = A.LocalWidth();
         const Real* ABuf = A.LockedBuffer();
         const Int ALDim = A.LDim();
+        for( Int jLoc=0; jLoc<nLocal; ++jLoc )
+            for( Int iLoc=0; iLoc<mLocal; ++iLoc )
+                value = Max(abs(value),ABuf[iLoc+jLoc*ALDim]);
```
poulson: Same comment on `s/std::abs/El::Abs/` here

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -11,46 +11,58 @@
...
-            for( Int jLoc=0; jLoc<nLocal; ++jLoc )
-                for( Int iLoc=0; iLoc<mLocal; ++iLoc )
-                    value = Max(std::abs(value),ABuf[iLoc+jLoc*ALDim]);
-
-            value = mpi::AllReduce( value, mpi::MAX, A.DistComm() );
-        }
-        mpi::Broadcast( value, A.Root(), A.CrossComm() );
-        return value;
-    }
+#define PROTO(Real) \
```
poulson: `s/Real/F/` here

src/blas_like/level1/MaxAbs.cpp
```diff
@@ -11,46 +11,58 @@
...
+  template Base<Real> MaxAbs( const Matrix<Real>& x ); \
+  template Base<Real> MaxAbs( const AbstractDistMatrix<Real>& x ); \
 
-}
+#define EL_NO_COMPLEX_PROTO
```
poulson: The `#define EL_NO_COMPLEX_PROTO` should be removed since this routine can trivially support complex inputs.

tests/blas_like/MaxAbs.cpp
```diff
@@ -0,0 +1,53 @@
...
+    assert( MaxAbs( A ) == Real(1) );
+    MinIJ( A, m );
+    assert( MaxAbs( A ) == Real(m) );
+    OneTwoOne( A, m );
+    assert( MaxAbs( A ) == Real(2) );
+}
+
+void RunTests( Int m)
+{
+  TestMaxAbs<double,DistMatrix<double>>( m );
```
poulson: Why suddenly only 2 spaces instead of 4?

tests/blas_like/MaxAbs.cpp
```diff
@@ -0,0 +1,53 @@
...
+int main( int argc, char* argv[] )
+{
+    Environment env( argc, argv );
+    mpi::Comm comm = mpi::COMM_WORLD;
+    try
+    {
+	Int m = 1;
+	for( Int e = 1; e<4; ++e )
+	{
+		m *= 10;
```
poulson: Ah, 4 spaces, then 2 spaces, then 6 spaces ;-)

davidsblom:
Thanks for all the feedback! Let me know if I there are further suggestions.

poulson:
By the way, at a minimum we should add your name to the files you changed and the AUTHORS file. Thank you for the contribution!

davidsblom:
That would be great. Thanks for helping out!

PR #170, opened 2016-08-27
Fix installed names

rhl-:
This cleans up the double negation.


poulson:
Thank you for humoring my pedantry! I will fix some of the spacing issues manually after merging to avoid any more of it since I agree that styling should be automated.

PR #171, opened 2016-08-27
installed catch unit testing framework.

rhl-:
I intend to use this (rather popular) unit testing framework within unit tests going forward. It is BSD licensed


poulson:
FWIW, it is under the Boost license according to the file header. It would be good to commit at least one nontrivial example in the PR.

rhl-:
closing this temporarily as I am realizing it may be nontrivial to integrate into existing unit tests.

PR #183, opened 2016-10-18
WIP: Add BFGS & Catch Test Framework

rhl-:
This Pull Request adds support for Basic BFGS within Elemental. I am using the Zoom line search algorithm found on Page 60-61 of Nocedal & Wright. The BFGS implementation itself maintains the updates to the inverse of the approximate hessian, and simply applies them at every iteration.

I have tested the optimizer with three functions, f(x) = x'x, and f(x) = x'A'Ax + x'b, and the rosenbrock function. As such I have growing confidence in the correctness of the procedure. 

The scheme uses random starting vectors, and I have extremely infrequently seen some of the tests converge to a vector of NaN's. I am not yet sure why this is happening, as I am not able to reproduce it regularly. Any help here would be appreciated.

Nocedal & Wright provide no guidance on the setting of alphaMax, which I have set to 100. There are a few parameters set which I am not confident on. Also, I would appreciated commentary on the stopping criteria for the Zoom procedure.

In addition, I have edited the CMakeLists.txt to allow the interactive CLion Debugger to function with Elemental, and I have removed the logic which forces a recompile on a git commit. 

With 30-60 minute long recompiles, this logic gets in the way of active development.


CMakeLists.txt
```diff
@@ -259,27 +259,23 @@ if(PROJECT_SOURCE_DIR STREQUAL PROJECT_BINARY_DIR)
...
-  message(WARNING "RelWithDebInfo not supported; switching to Release")
-  set(CMAKE_BUILD_TYPE Release)
-elseif(CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
-  message(WARNING "MinSizeRel not supported; switching to Release")
-  set(CMAKE_BUILD_TYPE Release)
-else()
-  message(WARNING "Build mode not specified, defaulting to Release build.")
-  set(CMAKE_BUILD_TYPE Release)
-endif()
+#if(CMAKE_BUILD_TYPE STREQUAL "Release")
```
poulson: What do you do when no CMake build type was specified?

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,212 @@
...
+*/
+#ifndef EL_OPTIMIZATION_BFGS_HPP
+#define EL_OPTIMIZATION_BFGS_HPP
+#include <utility>
+#include <El.hpp>
+
+namespace El {
+
+template< typename T, typename Function, typename Gradient>
+T zoom( const Function& f, const Gradient& gradient, T f0, const DistMatrix<T>& x0, const DistMatrix<T>& p,
```
poulson: Why the insistence on starting function names with lowercase variables (as opposed to the 10,000 other functions in the library)?

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,212 @@
...
+        T alpha_low, T alpha_high, T c1, T c2){
+    DistMatrix<T> x_j(x0);
+    DistMatrix<T> g2(p.Height(), 1);
+    while (alpha_high - alpha_low > 10*El::limits::Epsilon<T>()){
```
poulson: Is `T` assumed to be a real variable? Otherwise one should use `El::limits::Epsilon<Base<T>>()`

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,212 @@
...
+        T alpha = (alpha_low + alpha_high)/2.0;
```
poulson: I would highly recommend against ever using floating-point literals in templated routines and would instead recommend `T(2)` instead of `T(2.0)`. This happens to be safe for `2.0` but the conversion generally leads to a loss in precision.

include/El/optimization/bfgs.hpp
rhl-: Sorry :) I'm trying my best to do camel case.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,212 @@
...
+ * we are using this inside of a larger algorithm on an approximate problem, there is no reaosn to believe
+ * that this is the "best" step. Instead we just hope for a step with "sufficient" decrease, and then we
+ * hope to prove that if {x_i} approaches the minimum that this function converges.
+ * This algorithm attempts to satsify the weak wolf conditions.
+ */
+template< typename T, typename Function, typename Gradient>
+T lineSearch( const Function& f, const Gradient& gradient,
+              const DistMatrix<T>& g, Int D,
+              const DistMatrix<T>& x0, const DistMatrix<T>& p,
+              Int maxIter=100, T c1=1e-4, T c2=0.9){
```
rhl-: Can you recommend the specific function?

poulson: `Pow(limits::Epsilon<Real>(),Real(1)/Real(4))` would return the fourth-root of machine epsilon, which is roughly `1-e4` for double-precision. I'm not saying that this is necessarily the right generalization, but it's worth considering whether you want 1e-4 for single precision.

CMakeLists.txt
rhl-: Not sure what "CMake" mode this is, but, it appears that the output is no optimization flags on my mac. Maybe that means -O0 by default?

rhl-: I've gone ahead and added this check, and setting to Release in the event that its not set. In the event that this is not set, there are numerous build errors within Elemental. This seems like a deficiency. I'll see if I can look more into this.

rhl-: I am just trying to use the CLion IDE Interactive Debugger. With those lines it will not build in debug mode, without them, it will. 

What I said was a deficiency is that if you manage to leave CMAKE_BUILD_TYPE unspecified it gives a number of compile errors. I don't think there is anything controversial about that.

rhl-: If you would like I can leave this change out of the diff, and we can discuss it separately.

CMakeLists.txt
```diff
@@ -259,6 +259,9 @@ if(PROJECT_SOURCE_DIR STREQUAL PROJECT_BINARY_DIR)
   message(FATAL_ERROR "In-source build attempted; please clean the CMake cache and then switch to an out-of-source build, e.g.,\nrm CMakeCache.txt && rm -Rf CMakeFiles/\nmkdir build/ && cd build/ && cmake ..")
 endif()
 
+if (NOT CMAKE_BUILD_TYPE)
```
poulson: It is also important that we handle mis-set build types (e.g., the old `PureDebug` choice should cause an error). I think that we need to enumerate the allowed values and complain otherwise.

include/El/optimization/bfgs.hpp
```diff
@@ -47,18 +47,18 @@ template< typename T, typename Function, typename Gradient>
...
+        DistMatrix<T> g2(D, 1, x0.Grid());
         T f0_dash = Dot(p,g);
-        T  alpha = 1;
-        T  alpha_prev = 0;
-        T  alphaMax = 1e3;
-        T fvalPrev = 0;
-        T fval = 0;
+        T  alpha(1);
+        T  alpha_prev(0);
+        T  alphaMax(1e3);
```
poulson: I am not completely sure, but my intuition is that `1e3` is a double-precision literal.

include/El/optimization/bfgs.hpp
```diff
@@ -14,12 +14,12 @@
...
 template< typename T, typename Function, typename Gradient>
-T zoom( const Function& f, const Gradient& gradient, T f0, const DistMatrix<T>& x0, const DistMatrix<T>& p,
+T Zoom( const Function& f, const Gradient& gradient, T f0, const DistMatrix<T>& x0, const DistMatrix<T>& p,
         T alpha_low, T alpha_high, T c1, T c2){
     DistMatrix<T> x_j(x0);
-    DistMatrix<T> g2(p.Height(), 1);
-    while (alpha_high - alpha_low > 10*El::limits::Epsilon<T>()){
-        T alpha = (alpha_low + alpha_high)/2.0;
+    DistMatrix<T> g2(p.Height(), 1, x0.Grid());
+    while (alpha_high - alpha_low > T(10)*El::limits::Epsilon<Base<T>>()){
```
poulson: There is no reason to need the `El::` prefix in this case since this function is defined within the `El` namespace. The only exceptions that I'm aware of are due to conflicts caused by Argument Dependent Lookup where one of the arguments comes from a different namespace which has a function with the same name as the one from the `El` namespace.

poulson: Also, for the sake of consistency with the rest of the library, can we use symmetry with braces and put them on a line by themselves?

include/El/optimization/bfgs.hpp
```diff
@@ -47,18 +47,18 @@ template< typename T, typename Function, typename Gradient>
 T lineSearch( const Function& f, const Gradient& gradient,
```
rhl-: done.

include/El/optimization/bfgs.hpp
poulson: It is moreso to call attention to the fact that there is a conversion happening. This is fine in this case since double-precision represents 1e3 exactly.

include/El/optimization/bfgs.hpp
```diff
@@ -176,12 +176,12 @@ T BFGS( Vector& x, const std::function< T(const Vector&)>& f,
     gradient(x, g);
     detail::HessianInverseOperator<T> Hinv;
     auto norm_g = InfinityNorm(g);
-    for( std::size_t iter=0; (norm_g > 100*limits::Epsilon<T>()); ++iter){
+    for( std::size_t iter=0; (norm_g > T(100)*limits::Epsilon<Base<T>>()); ++iter){
         //std::cout << "iter: " << iter << std::endl;
         //El::Display(x, "Iterate");
```
poulson: Can we remove the debugging artifacts?

CMakeLists.txt
rhl-: Is that an old Elemental build type? whats the difference between it and Debug?

CMakeLists.txt
```diff
@@ -260,7 +260,9 @@ if(PROJECT_SOURCE_DIR STREQUAL PROJECT_BINARY_DIR)
 endif()
 
 if (NOT CMAKE_BUILD_TYPE)
-    set( CMAKE_BUILD_TYPE Release)
+   set( CMAKE_BUILD_TYPE Release)
+elseif(CMAKE_BUILD_TYPE STREQUAL "PureDebug")
```
poulson: This only handles one particular case. Any misspelled or non-standard build type will fail. It would be better to warn and switch to one of the standard ones (e.g., Release).

rhl-: I'm not sure why we need to do any of this. CMake has a list of supported build types. I've never heard of PureDebug, and I don't know why anyone would try to enter it.

> Possible values are empty, Debug, Release, RelWithDebInfo and MinSizeRel.

From: https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html

rhl-: What about adding this: http://public.kitware.com/pipermail/cmake/2014-June/057852.html 

perhaps with a call to the uppercase as the author suggests?

poulson: Elemental used to use `{Pure,Hybrid}{Debug,Release}` to specify whether multithreading was expected and whether debug or release (and the ill-advised `EL_HYBRID` option is now separate). The point is that, if there are only four options, it makes sense to complain as early as possible if Elemental breaks if none of those four are used. (And we need to make sure that `MinSizeRel` and `RelWithDebInfo` both set `EL_RELEASE` to `ON`)

rhl-: Ah, that makes more sense to me. I didn't understand that. I think the solution I suggested should be fine. Let me take a crack at it.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+            {
+                return alpha;
+            }
+            if( fval_dash > 0)
+            {
+                return Zoom(f, gradient, f0, x0, p, alpha, alpha_prev, c1, c2);
+            }
+            alpha_prev = alpha;
+            fvalPrev = fval;
+            alpha = std::min(2*alpha, alphaMax);
```
poulson: I'm not sure that `std::min` works for nonstandard datatypes (e.g. `El::BigFloat`) and it would be better to use Elemental's `Min`.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+}
+
+namespace detail {
+/***
+ * This class stores the updates to the inverse hessian of F, and provides the ability to apply inverse hessian.
+ */
+template< typename T>
+struct HessianInverseOperator {
+
+    typedef DistMatrix<T> Matrix;
```
poulson: *Please* do not typedef `El::DistMatrix` to `El::Matrix` when `El::Matrix` already exists...

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+        // We store s,y, alpha, H*y, and s'y.
+        auto sy = Dot(s,y);
+        auto Hy = (*this)*y;
+        auto yHy = Dot(Hy,y);
+        T alpha = (sy + yHy)/(sy*sy);
+        _hessian_data.emplace_back( alpha, s, y, Hy, sy);
+    }
+
+private:
+    Hessian_updates _hessian_data;
```
poulson: It is usually considered a bad idea to prefix an underscore, as variables which start with an underscore and are followed by a capital letter are reserved for internal implementations (and I have been burned by this in the past). For this reason, it is better to postfix the underscore to avoid later problems.

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,172 @@
...
+  const std::function< T(const DistMatrix<T>&)>
+  logistic_function = [&](const DistMatrix<T>& theta){
+     auto nrm = El::Norm(theta);
+     El::DistMatrix<T> y(theta);
+     El::DistMatrix<T> r(X.Height(), 1);
+     El::Gemv(El::NORMAL, T(-1), X, theta, T(0), r);
+     for( auto i = 0; i < y.Height(); ++i){ r.Set(i, 0, y.Get(i,0));  }
+
+     T v = lambda*nrm*nrm;
+     for(auto i = 0; i < y.Height(); ++i){ v += El::Log(T(1) + El::Exp(-1*y.Get(i,0)*r.Get(i,0))); }
```
poulson: This is going to be monstrously slow due to (1) being linear runtime due to every process performing linear work, and (2) due to there being `2*y.Height()` broadcasts within the `Get` calls. It would be better to locally perform the sums and then to reduce at the end with a single `mpi::AllReduce`.

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,172 @@
...
+ * @return
+ */
+template< typename T>
+std::pair< DistMatrix<T>, T>
+QuadraticFunction( const DistMatrix<T> & A, const DistMatrix<T>& b){
+  const std::function< T(const DistMatrix<T>&)>
+  quadratic_function = [&](const DistMatrix<T>& theta){
+        DistMatrix<T> y( theta);
+        Gemv(El::NORMAL, T(1), A, theta, T(0), y);
+        return .5*(Dot(y,theta) - Dot(b, theta));
```
poulson: The usage of `.5` here is ill-advised due to it being a double-precision literal. It would be better to divide the difference of the dot products by `T(2)`. (And I would prefer `F` is the type is assumed to be a field, which it is here due to the division)

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,172 @@
...
+  auto val = El::BFGS( x0, quadratic_function,  gradient);
+  return std::make_pair( x0, val);
+}
+
+template< typename T>
+std::pair< DistMatrix<T>, T>
+QuadraticBFGSTest(Int N){
+ DistMatrix<T> A,B;
+ Laplacian(A, N);
+ auto nrm = Norm(A);
```
poulson: `Norm` defaults to the two-norm, which involves computing the SVD. Perhaps you want `TwoNormEstimate` or `FrobeniusNorm`?

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,172 @@
...
+ Gemm(El::NORMAL, El::ADJOINT, T(1), A, A, B);
+ A = B;
+ A *= T(1)/(nrm*nrm);
+ DistMatrix<T> b( N, 1);
+ Gaussian(b, N, 1);
+ b *= (T(1)/Norm(b));
```
poulson: Same here. `FrobeniusNorm` is probably what you want.

tests/optimization/bfgs.cpp
rhl-: I'm going to temporarily remove this and bring it back later. I have changed this on a separate branch. It's still not working perfectly, and I'm not confident I have the right formulae. I'm interested in just appreciating that the core BFGS procedure is satisfactory.

include/El/optimization/bfgs.hpp
poulson: They are also often (whether improperly or not) reserved by macros. There was a MPI profiling framework (I believe tau?) which did so and caused problems for me in the past.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+                return alpha;
+            }
+            if( fval_dash > 0)
+            {
+                return Zoom(f, gradient, f0, x0, p, alpha, alpha_prev, c1, c2);
+            }
+            alpha_prev = alpha;
+            fvalPrev = fval;
+            alpha = Min(2*alpha, alphaMax);
+        }
```
poulson: The indentation seems rather strange here.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+namespace detail {
+/***
+ * This class stores the updates to the inverse hessian of F, and provides the ability to apply inverse hessian.
+ */
+template< typename T>
+struct HessianInverseOperator {
+
+    typedef DistMatrix<T> DMatrix;
+    typedef std::tuple< T, DMatrix, DMatrix, DMatrix, T> Update;
+    typedef std::vector< Update > Hessian_updates;
```
poulson: The angle bracket spacing seems to be varying line-by-line. The majority of the library uses no spaces after `<` or before `>`, except for multi-line alignments in some cases.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+     * where rho = (y'z)/(s'y)
+     *
+     * @param x
+     * @return H_k*x
+     */
+    DMatrix operator*( const DMatrix& x)
+    {
+        //Initially this is just the identity matrix;
+        //H_0 = I, so H_0*x = x;
+        if (_hessian_data.size() == 0) { return x; }
```
poulson: As mentioned before, it is usually considered bad form to have member variables that begin with `_` (as post-fixes are preferred).

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+     * The forward update is:
+     * B_{k+1} = B_k + yy'/(y's) + B_ksks'B_k/(s'Bs)
+     * So the backwards update is:
+     * H_{k+1} = H_k + alpha*(ss') - [H_kys' + sy'H_k]/(s'y)
+     * where alpha = [(s'y + y'H_ky)/(s'y)^2];
+     * This method advances the operator to represent H_{k+1}
+     * @param s
+     * @param y
+     */
+    void update( DMatrix& s, DMatrix& y)
```
poulson: It would be more stylistically consistent to call this routine `Update`.

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,149 @@
...
+   which can be found in the LICENSE file in the root directory, or at 
+   http://opensource.org/licenses/BSD-2-Clause
+*/
+#include <utility>
+#include <El.hpp>
+#include <El/optimization/bfgs.hpp>
+#define CATCH_CONFIG_RUNNER
+#include <catch.hpp>
+
+using namespace El;
```
poulson: There is a rather strange mix of using and not using the `El::` prefix for member variables and constants below.

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+template< typename T>
+std::pair< DistMatrix<T>, T>
+SimpleQuadraticBFGSTest( const Int & N){
+  const std::function< T(const DistMatrix<T>&)>
```
poulson: Why the switch to two-space indentation here?

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,149 @@
...
+ * @return
+ */
+template< typename T>
+std::pair< DistMatrix<T>, T>
+QuadraticFunction( const DistMatrix<T> & A, const DistMatrix<T>& b)
+{
+  const std::function< T(const DistMatrix<T>&)>
+  quadratic_function = [&](const DistMatrix<T>& theta)
+  {
+        DistMatrix<T> y( theta);
```
poulson: Two-space followed by six-space?

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,149 @@
...
+ A *= T(1)/(nrm*nrm);
+ DistMatrix<T> b( N, 1);
+ Gaussian(b, N, 1);
+ b *= (T(1)/FrobeniusNorm(b));
+ return QuadraticFunction<T>(A, b);
+}
+
+template< typename T>
+std::pair< DistMatrix<T>, T>
+RosenbrockTest(){
```
poulson: Thank you for adding a Rosenbrock test!

tests/optimization/bfgs.cpp
```diff
@@ -0,0 +1,149 @@
...
+    const std::function< T(const DistMatrix<T>&)>
+          rosenbrock = [&](const DistMatrix<T>& theta)
+    {
+        auto x1 = theta.Get(0,0);
```
poulson: Each `Get` call to a `DistMatrix` will involve an `mpi::Broadcast` under the hood; this is expensive for querying a vector of length two (when the vector could be duplicated across all processes), but I suppose this is okay for testing.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
+/*
+   Copyright (c) 2009-2016, Ryan H. Lewis
```
rhl-: neither am I. Fixed.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+*/
+#ifndef EL_OPTIMIZATION_BFGS_HPP
+#define EL_OPTIMIZATION_BFGS_HPP
+#include <utility>
+#include <El.hpp>
+
+namespace El {
+
+template< typename T, typename Function, typename Gradient>
+T Zoom( const Function& f, const Gradient& gradient, T f0, const DistMatrix<T>& x0, const DistMatrix<T>& p,
```
rhl-: These parameters match the notation in nocedal and wright. I'd prefer to leave them and mention that explicitly. Is that acceptable?

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+        }
+        gradient( x_j, g2);
+        auto fval_dash = Dot(p, g2);
+        if(  fval_dash*(alpha_high - alpha_low) >= 0)
+        {
+            alpha_high = alpha_low;
+        }
+        alpha_low = alpha;
+    }
+    return (alpha_high+alpha_low)/T(2.0);
```
rhl-: Fixed.

include/El/optimization/bfgs.hpp
rhl-: fixed

include/El/optimization/bfgs.hpp
rhl-: fixed

tests/optimization/bfgs.cpp
rhl-: :)

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,226 @@
...
+/***
+ * This function approximately minimizes the function min_\alpha f(x0 + alpha*p)
+ * In some sense minimizing this function is "ideal" however, this may take a lot of work, and since
+ * we are using this inside of a larger algorithm on an approximate problem, there is no reason to believe
+ * that this is the "best" step. Instead we just hope for a step with "sufficient" decrease, and then we
+ * hope to prove that if {x_i} approaches the minimum that this function converges.
+ * This algorithm attempts to satsify the weak wolf conditions.
+ */
+template< typename T, typename Function, typename Gradient>
+T lineSearch( const Function& f, const Gradient& gradient,
```
rhl-: I'm not sure what this line search is called. It's the one from nocedal and wright, more or less, but they don't name it. feel free to propose a name and i'll adopt it.

include/El/optimization/bfgs.hpp
rhl-: fixed

tests/optimization/bfgs.cpp
rhl-: see above.

tests/optimization/bfgs.cpp
rhl-: I try to always use namespaces (renaming them locally when they are too long to type). it enhances readability, but, I did copy this cpp file from another one in El.

tests/optimization/bfgs.cpp
rhl-: yeah, that's bad. How do I fix this again?

poulson: If `theta` is distributed as a `DistMatrix<T,STAR,STAR>` then one could call `GetLocal` instead.

include/El/optimization/bfgs.hpp
```diff
@@ -0,0 +1,423 @@
...
+/**
+ * Preconditions:
+ * -------------
+ * 1. The interval bounded by alphaLow and alphaHigh contains a value alpha satisfying the strong wolfe conditions.
+ * 2. alphaLow satisfies the sufficient decrease condition.
+ * 3. f(x + alphaLow*p)*(alphaHigh - alphaLow) < 0
+ *
+ * Note: it is _not_ required that alphaLow < alphaHigh.
+ *
+ * @param f
```
poulson: I would strongly recommend documenting what each routine is meant to accomplish before documenting the parameters.

include/El/optimization/bfgs.hpp
```diff
@@ -172,6 +172,7 @@ T lineSearch( const std::function< T(const Vector&)>& f, const Gradient& gradien
             }
         }while( !done);
         if( !limits::IsFinite(beta)){ RuntimeError("Line search failed to brack point satisfying weak wolfe conditions. Function may be unbounded below"); }
+        RuntimeError(std::string("[")+alpha+","+beta+"] brackets an interval containing a point satisfying WWC");
```
rhl-: cool.

PR #190, opened 2016-10-29
Dist hess schur

poulson:
Adding preliminary support for distributed (real and complex; multi-bulge and AED) Hessenberg QR algorithms. These algorithms have undergone modest testing and have not yet been performance tuned.

### DIFF
--- a/include/El/blas_like/level1/Copy/GeneralPurpose.hpp
+++ b/include/El/blas_like/level1/Copy/GeneralPurpose.hpp
@@ -205,7 +205,9 @@ void GeneralPurpose
     const bool useBLACSRedist = true;
     if( useBLACSRedist &&
         A.ColDist() == MC && A.RowDist() == MR &&
-        B.ColDist() == MC && B.RowDist() == MR )
+        B.ColDist() == MC && B.RowDist() == MR &&
+        A.ColCut() == 0 && A.RowCut() == 0 &&
+        B.ColCut() == 0 && B.RowCut() == 0 )
     {
         B.Resize( height, width );
         const int contextA = blacs::Context( A );
--- a/include/El/blas_like/level1/Copy/GeneralPurpose.hpp
+++ b/include/El/blas_like/level1/Copy/GeneralPurpose.hpp
@@ -27,6 +27,7 @@ void Helper
     B.Resize( height, width );
     Zero( B );
     const bool BPartic = B.Participating();
+    const int BRoot = B.Root();
 
     const bool includeViewers = (A.Grid() != B.Grid());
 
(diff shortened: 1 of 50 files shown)


include/El/blas_like/level1/Copy/GeneralPurpose.hpp
```diff
@@ -93,19 +98,24 @@ void Helper
     if( includeViewers )
     {
         comm = g.ViewingComm();
-        const int commSize = mpi::Size( comm );
+        const int viewingSize = mpi::Size( g.ViewingComm() );
+        const int distBSize = mpi::Size( B.DistComm() );
 
-        vector<int> distMap(commSize);
-        for( int q=0; q<commSize; ++q )
+        vector<int> distBToViewing(distBSize);
```
rhl-: std::vector<int>

include/El/blas_like/level1/Copy/GeneralPurpose.hpp
```diff
@@ -93,19 +98,24 @@ void Helper
...
+        for( int distBRank=0; distBRank<distBSize; ++distBRank )
```
rhl-: why `int` and not `Int`

src/core/DistMatrix/Abstract.cpp
```diff
@@ -582,42 +582,43 @@ void AbstractDistMatrix<T>::ProcessQueues( bool includeViewers )
...
+    // We will first push to redundant rank 0
+    const int redundantRoot = 0;
+
     // Compute the metadata
     // ====================
     mpi::Comm comm;
     vector<int> sendCounts, owners(totalSend);
     if( includeViewers )
     {
         comm = g.ViewingComm();
```
rhl-: same.

include/El/blas_like/level1/Copy/GeneralPurpose.hpp
poulson: Actually, Elemental imports `std::vector` as `El::vector` here: https://github.com/elemental/Elemental/blob/1bd259d51aa1e7f7ed40306580a302640c36b2d6/include/El/core/environment/decl.hpp#L19

include/El/blas_like/level1/Copy/GeneralPurpose.hpp
poulson: MPI forces ranks to be indexed with `int`, so this is actually correct

poulson:
I am going forward and merging since this fixes several outstanding bugs in `DistMatrix<T,MC,MR,BLOCK>`

PR #192, opened 2016-11-15
Adds soversioning to the El library.

rhl-:
This is per the review of the Elemental RPM.


rhl-:
The goal is to see that the build passes and things still work. not sure what this does.

poulson:
LGTM.

PR #196, opened 2016-11-19
PMRRR calls exit

rhl-:
This PR addresses another linter issue.

### DIFF
--- a/external/pmrrr/src/plarre.c
+++ b/external/pmrrr/src/plarre.c
@@ -697,7 +697,6 @@ int eigval_root_proc
       } else if (jtry == MAX_TRY_RRR-1) {
 	fprintf(stderr,"No initial representation could be found.\n");
 	return 3;
-	//exit(3);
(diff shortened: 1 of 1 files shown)


external/pmrrr/src/plarre.c
```diff
@@ -696,7 +696,8 @@ int eigval_root_proc
 	}
       } else if (jtry == MAX_TRY_RRR-1) {
 	fprintf(stderr,"No initial representation could be found.\n");
-	exit(3);
+	return 3;
+	//exit(3);
```
poulson: Upon further review, modulo removing the commented out line, I approve this PR, as the only callsite asserts that the return value is zero.

rhl-: comment removed.

poulson:
It is a good idea to CC @rocanale and @pauldj since they likely want to make the equivalent changes upstream as well.

PR #197, opened 2016-11-19
Add explicit libm dependency on linux

rhl-:
It appears that SuiteSparse on linux could explicitly depend on libm.


poulson:
Why only linux? And have you tested this yet?

rhl-:
We could remove the flags. the issue is only appearing on fedoras linter, so I figure its more specifically a linux issue. This would likely appear in a debian linter as well. 

I am testing it by issuing this PR, the travis build would test a debian failure. if it works there it will likely work on fedora, the copr build will test that after a merge.

PR #202, opened 2016-11-25
Fixed an example not building on 64-bit Ints.

AidanGG:
When not building on 64-bit `El::Ints`, the `0` literal is an `int`, which is of the same type as `El::Int`, but runs into issues when they are different types when using 64-bit `El::Ints` through `-DEL_USE_64BIT_BLAS_INTS=ON`. This fixes it with a simple cast.


poulson:
Thanks for taking the time to submit this!

PR #204, opened 2016-11-29
Adding missing sequential dense equilibration implementations

poulson:
The sequential dense LPs were non-functioning due to the sequential dense equilibration routines not having been implemented (despite the *distributed* dense implementations existing for some time, as well as the sequential and distributed sparse implementations).


rhl-:
You should consider use git cherry-pick in the future as the release branches and the master branch can diverge. For example, I just committed something to master.

PR #205, opened 2016-11-29
fix FindGMP.cmake

rhl-:
This correctly determines the GMP version at least when GCC is used as a compiler. I'm not sure what happens with clang or any more exotic compiler, such as icc or vendor compilers.

The main thing is the flags: -dM -E


poulson:
Why not only perform the new approach of the compiler is already recognized as gcc?

rhl-:
it looks like this works with clang and gcc, and googling now suggests it works with icc. It seems likes most compilers support this, so, lets stick with it.

PR #213, opened 2016-12-19
Replacing hardcoded 'lib' by the CMAKE variable in ElVars file

rocanale:
Now linking works when system uses 'lib64' or other variants


poulson:
Thanks!

PR #214, opened 2016-12-28
Adding encapsulated interfaces to the sparse LDL factorizations

poulson:


poulson:
Several bugs have now been fixed, an MPS format reader has been introduced, and the stability of the affine LP IPM is now improved via scaling similar to the direct case. The Interior Point Methods still deserve significantly more refactoring and rescaling attention.

PR #217, opened 2017-01-29
Fixed default grid height calculation.

AidanGG:
For example, a four process grid was 4x1, when it should have been 2x2.


poulson:
Good catch. I must have transposed the order in a recent refactoring!

PR #221, opened 2017-02-18
Parallelizing several BLAS 1 (-like) routines with OpenMP

timmoon10:
Minimal changes are required since it uses the "EL_PARALLEL_FOR_COLLAPSE2" macro.

This commit will break the behavior of EntrywiseMap, IndexDependentFill, and IndexDependentMap if the user-provided function is not thread-safe. I hold off of parallelizing EntrywiseFill since Uniform and other random matrices are generated by passing a static pseudorandom number generator to EntrywiseFill.

### DIFF
--- a/cmake/configure_files/config.h.in
+++ b/cmake/configure_files/config.h.in
@@ -79,6 +79,7 @@
 #cmakedefine EL_HAVE_PRETTY_FUNCTION
 #cmakedefine EL_HAVE_OPENMP
 #cmakedefine EL_HAVE_OMP_COLLAPSE
+#cmakedefine EL_HAVE_OMP_SIMD
 #cmakedefine EL_HAVE_QT5
 #cmakedefine EL_AVOID_COMPLEX_MPI
 #cmakedefine EL_HAVE_CXX11RANDOM
--- a/cmake/detect/OpenMP.cmake
+++ b/cmake/detect/OpenMP.cmake
@@ -46,3 +46,22 @@ if(EL_HAVE_OPENMP)
 else()
   set(EL_HAVE_OMP_COLLAPSE FALSE)
 endif()
+
+# See if we have 'simd' support, which was introduced in OpenMP 4.0
+if(EL_HAVE_OPENMP)
+  set(CMAKE_REQUIRED_FLAGS ${OpenMP_CXX_FLAGS})
+  set(OMP_SIMD_CODE
+      "#include <omp.h>
+       int main( int argc, char* argv[] ) 
+       {
+           int k[10];
+       #pragma omp simd
+           for( int i=0; i<10; ++i )
+               k[i] = i;
+           return 0; 
+       }")
+  check_cxx_source_compiles("${OMP_SIMD_CODE}" EL_HAVE_OMP_SIMD)
+  set(CMAKE_REQUIRED_FLAGS)
+else()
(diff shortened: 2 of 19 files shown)


poulson:
Thanks Tim! Is there a particular benchmark that you are using to test the impact of these changes?

timmoon10:
I haven't benchmarked this change, although it would be a good idea. I can write something quick for EntrywiseMap.

jeffhammond:
I don't know if it is worth trying to address, but `collapse(n)` has an unfortunate effect of requiring modulo and division computations in loops in some (many?) cases.  This hurts performance since these are expensive instructions.

The recommended pattern for double loops is `omp for` (or `omp parallel for` if not already in parallel) on the outer loop and `omp simd` on the inner loop.

poulson:
@jeffhammond These types of issues are why I think that any performance-motivated multithreaded development should be coupled with experiments on at least one architecture that demonstrate speedups over the sequential analogue.

jeffhammond:
Yeah, the other thing to do is probably not bother to thread when the loop trips are small.  I'm not aware of any OpenMP implementation that determines whether or not it is worthwhile to thread and runs in serial if the parallel overhead isn't worth it (because this is, in general, an impossible problem).

timmoon10:
I think using `omp simd` is a good idea for simple routines like Fill or Hadamard. However, vectorization may run into trouble in EntrywiseMap, IndexDependentMap, and IndexDependentFill if the user provides a function with conditional branching.

timmoon10:
I've benchmarked EntrywiseMap with the serial implementation, the `collapse(2)` implementation, and the `for`/`simd` implementation. These experiments were performed with single-precision exponential functions computed on two 12-core Intel Xeon E5-2695v2 processors:
slide1
slide2
We see that the performance of the OpenMP implementations scales with the number of OpenMP threads (until we reach the CPU core count). We also see that the `for`/`simd` implementation outperforms the serial and `collapse(2)` implementations at every matrix size. In my experiments with an unbalanced function (the number of flops is drawn from a power-law distribution), the `for`/`simd` implementation is also comparable/better than the others.

jeffhammond:
I assume the 24-32-48 inversion is due to crossing the socket boundary.  What affinity settings are you using?  This is mostly curiosity not a suggestion that there is anything wrong with the code.

timmoon10:
That sounds like the case. I see a big performance hit when I go from 24 processes to 25. Here are my settings:
```
~/experiments/scripts $ srun printenv | grep OMP
LMOD_FAMILY_COMPILER=intel
TACC_FAMILY_COMPILER=intel
OMP_NUM_THREADS=24
GOMP_CPU_AFFINITY=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23
```

jeffhammond:
If you are using the Intel compiler, I recommend `KMP_AFFINITY=granularity=fine,scatter` and `KMP_AFFINITY=granularity=fine,compact` rather than trying to specify processor affinity via enumeration.

If Intel-specific options are undesirable, `OMP_PROC_BIND=SPREAD` and `OMP_PROC_BIND=CLOSE` have a similar effect.

I suspect that using breadth-first (scatter or spread) affinity will smooth out the performance, although perhaps reduce performance in cases where the working set fits into cache.

Issue #224, opened 2017-03-21
Block matrices in BLAS methods


ndryden:
It looks like several of the BLAS methods currently only support element-wise matrices as arguments, and do not support block-wise matrices.

Functions I particularly care about are:
* `Hadamard`
* `Dot` (which implies support in `HilbertSchmidt`)
* `ColumnTwoNorms`
* `ColumnMaxNorms`

I haven't looked, but I suspect there are more that don't support it that I'm not using right now.

poulson:
Many of these don't exist for legacy reasons (when they were originally written, `El::DistMatrix<T,colDist,rowDist,El::BLOCK>` did not yet exist), but each should only require a few lines of code to add support for.

Might I ask why you are preferring the blocked distribution? It was primarily meant for implementing the distributed Hessenberg QR algorithm and for legacy interfaces to ScaLAPACK.

ndryden:
Yes, it didn't look too hard to make the changes. I'll try and put together a pull request adding support for these methods.

The use case is for better distributed GEMM performance in code where the matrix multiplies are a major bottleneck (training neural networks).

poulson:
But, why use block distributions for the distributed GEMM? The element-wise distributions will have better performance in Elemental. There are a few legitimate reasons to demand block distributions (e.g., for bulge-chasing algorithms like the Hessenberg QR algorithm), but element-wise distributions are much simpler and block distributions are not quite first-class in the library (as you are finding).

ndryden:
I benchmarked this some time ago and found the opposite conclusion-- that block matrices ended up quite a bit faster, and didn't think it unusual (since it agrees with what I recall from theory). However, I just rewrote and ran a quick benchmark and the results agree with you: the element-wise distribution is faster. (Seems to be about 60-75% faster.) Which is confusing, but perhaps there was an issue with the prior benchmark or how I ran it.

Is the performance difference due to something inherent in the element-wise distribution, or just Elemental's distributed GEMM implementations?

poulson:
Despite popular opinion, there is nothing about element-wise distributions that effects the performance of GEMM relative to a block distribution, as each process stores its portion of matrices locally and can make use of BLAS 3. Further, nothing would prevent an implicit permutation of the rows and columns so that one could run the same algorithm as for the blocked case on an implicit permutation (though this doesn't hold for factorizations because, for example, triangular structure is not preserved under such permutations).

[...]

PR #226, opened 2017-03-24
Support block matrices in several BLAS1 routines

ndryden:
This adds support for block matrices in the `Hadamard`, `Dot`, `HilbertSchmidt`, `ColumnTwoNorms`, and `ColumnMaxNorms` functions (see #224). I also added tests for `Hadamard`, `Dot`, `ColumnTwoNorms`, and `ColumnMaxNorms`.

Let me know if any additional changes are needed.

### DIFF
--- a/include/El/blas_like/level1/Hadamard.hpp
+++ b/include/El/blas_like/level1/Hadamard.hpp
@@ -42,9 +42,9 @@ void Hadamard( const Matrix<T>& A, const Matrix<T>& B, Matrix<T>& C )
 
 template<typename T>
 void Hadamard
-( const ElementalMatrix<T>& A,
-  const ElementalMatrix<T>& B,
-        ElementalMatrix<T>& C )
+( const AbstractDistMatrix<T>& A,
+  const AbstractDistMatrix<T>& B,
+        AbstractDistMatrix<T>& C )
 {
     EL_DEBUG_CSE
     const DistData& ADistData = A.DistData();
@@ -87,9 +87,9 @@ void Hadamard
   EL_EXTERN template void Hadamard \
   ( const Matrix<T>& A, const Matrix<T>& B, Matrix<T>& C ); \
   EL_EXTERN template void Hadamard \
-  ( const ElementalMatrix<T>& A, \
-    const ElementalMatrix<T>& B, \
-          ElementalMatrix<T>& C ); \
+  ( const AbstractDistMatrix<T>& A, \
+    const AbstractDistMatrix<T>& B, \
+          AbstractDistMatrix<T>& C ); \
   EL_EXTERN template void Hadamard \
   ( const DistMultiVec<T>& A, \
     const DistMultiVec<T>& B, \
--- a/include/El/blas_like/level1/Hadamard.hpp
+++ b/include/El/blas_like/level1/Hadamard.hpp
@@ -60,6 +60,9 @@ void Hadamard
         LogicError("A, B, and C must share the same distribution");
     if( A.ColAlign() != B.ColAlign() || A.RowAlign() != B.RowAlign() )
         LogicError("A and B must be aligned");
+    if ( A.BlockHeight() != B.BlockHeight() ||
+         A.BlockWidth() != B.BlockWidth())
+      LogicError("A and B must have the same block size");
     C.AlignWith( A.DistData() );
     C.Resize( A.Height(), A.Width() );
     Hadamard( A.LockedMatrix(), B.LockedMatrix(), C.Matrix() );
--- a/tests/blas_like/ColumnNorms.cpp
+++ b/tests/blas_like/ColumnNorms.cpp
@@ -33,7 +33,10 @@ void TestColumnTwoNorms(Int m, Int n, const Grid& g, bool print)
     }
     expected = mpi::AllReduce(expected, g.ColComm());
     expected = Sqrt(expected);
-    if (Abs(got - expected) > 10 * limits::Epsilon<El::Base<T>>())
+    // Compute max(expected, 1) to use relative bound.
+    // (std::max and El::Max don't support BigFloat.
+    T div = expected > 1 ? expected : 1;
+    if (Abs(got - expected) / div > m * n * 10 * limits::Epsilon<El::Base<T>>())
     {
       Output("Results do not match, norms(", j, ")=", got,
              " instead of ", expected);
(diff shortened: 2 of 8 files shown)


tests/blas_like/ColumnNorms.cpp
```diff
@@ -0,0 +1,97 @@
...
+  ColumnTwoNorms(A, norms);
+  if (print) {
+    Print(norms, "norms");
+  }
+  for (Int j = 0; j < A.LocalWidth(); ++j) {
+    T got = norms.GetLocal(j, 0);
+    T expected = 0;
+    for (Int i = 0; i < A.LocalHeight(); ++i) {
+      T val = A.GetLocal(i, j);
+      expected += val * val;
```
poulson: Explicitly accumulating the squares and then square-rooting should be fine for random data but can lead to very low accuracy in extreme cases. There is the routine `El::UpdateScaledSquare` for accumulating the square of the two norm as the product of a scale and a scaled square that should generally be preferred (and which is used within the norm computation routines).

tests/blas_like/ColumnNorms.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+    expected = mpi::AllReduce(expected, g.ColComm());
+    expected = Sqrt(expected);
+    if (Abs(got - expected) > 1e-5) {
```
poulson: It would be more rigorous to use a bound of the form `n * limits::Epsilon<El::Base<T>>()` times a small constant (e.g., 10).

tests/blas_like/Dot.cpp
```diff
@@ -0,0 +1,63 @@
...
+    const bool print = Input("--print", "print matrices?", false);
+    ProcessInput();
+    PrintInputReport();
+
+    const Grid g(comm);
+    OutputFromRoot(comm, "Testing Dot");
+    TestDot<float, ELEMENT>(m, n, g, print);
+    TestDot<float, BLOCK>(m, n, g, print);
+    TestDot<double, ELEMENT>(m, n, g, print);
+    TestDot<double, BLOCK>(m, n, g, print);
```
poulson: Why exclude complex tests and more precise datatypes (e.g., `El::DoubleDouble`, `El::QuadDouble`, `El::Quad`, and `El::BigFloat`)? I would be happy to add the appropriate flag guards for the higher-precision tests, but the complex arithmetic variants also working is crucial.

tests/blas_like/Dot.cpp
```diff
@@ -0,0 +1,63 @@
...
+  } catch (exception& e) {
```
poulson: Nit: Elemental has made the (perhaps dubious) choice of putting curly braces on their own line.

tests/blas_like/Hadamard.cpp
```diff
@@ -0,0 +1,64 @@
...
+  mpi::Barrier(g.Comm());
+  if (print) {
+    Print(C, "C");
+  }
+  // Manually check results.
+  for (Int j = 0; j < A.LocalWidth(); ++j) {
+    for (Int i = 0; i < A.LocalHeight(); ++i) {
+      T got = C.GetLocal(i, j);
+      T expected = A.GetLocal(i, j) * B.GetLocal(i, j);
+      if (Abs(got - expected) > 1e-6) {
```
poulson: Same comment here about using `El::limits::Epsilon<El::Base<T>>()` times a small constant.

tests/blas_like/Hadamard.cpp
```diff
@@ -0,0 +1,64 @@
...
+    const bool print = Input("--print", "print matrices?", false);
+    ProcessInput();
+    PrintInputReport();
+
+    const Grid g(comm);
+    OutputFromRoot(comm, "Testing Hadamard");
+    TestHadamard<float, ELEMENT>(m, n, g, print);
+    TestHadamard<float, BLOCK>(m, n, g, print);
+    TestHadamard<double, ELEMENT>(m, n, g, print);
+    TestHadamard<double, BLOCK>(m, n, g, print);
```
poulson: Same comment here about testing the complex (and high-precision) cases.

include/El/blas_like/level1/Hadamard.hpp
```diff
@@ -60,6 +60,9 @@ void Hadamard
         LogicError("A, B, and C must share the same distribution");
     if( A.ColAlign() != B.ColAlign() || A.RowAlign() != B.RowAlign() )
         LogicError("A and B must be aligned");
+    if ( A.BlockHeight() != B.BlockHeight() ||
```
poulson: Thank you noticing this crucial portion of the extension!

tests/blas_like/ColumnNorms.cpp
ndryden: My main goal there was to provide a simple baseline that ensures there's no major issues (e.g. block vs element distributions causing problems) while not just reimplementing the method that's being tested. If you think it would be better to switch to `UpdateScaledSquare` in this case, I can do that.

ndryden:
My latest commit addresses your comments (except see the discussion on `UpdateScaledSquare`). Let me know if there's any other changes required!

tests/blas_like/Dot.cpp
```diff
@@ -55,9 +59,39 @@ int main(int argc, char** argv) {
...
     TestDot<float, ELEMENT>(m, n, g, print);
     TestDot<float, BLOCK>(m, n, g, print);
+    TestDot<Complex<float>, ELEMENT>(m, n, g, print);
+    TestDot<Complex<float>, BLOCK>(m, n, g, print);
     TestDot<double, ELEMENT>(m, n, g, print);
     TestDot<double, BLOCK>(m, n, g, print);
-  } catch (exception& e) {
+    TestDot<Complex<double>, ELEMENT>(m, n, g, print);
+    TestDot<Complex<double>, BLOCK>(m, n, g, print);
+#if defined(EL_HAVE_QD) && defined(EL_ENABLE_DOUBLEDOUBLE)
```
poulson: There is no need for the EL_ENABLE macros here (or in any test driver), as they are only used to signal to some of the include macros used for template instantiation in the library.

tests/blas_like/ColumnNorms.cpp
```diff
@@ -11,28 +11,30 @@
...
-    for (Int i = 0; i < A.LocalHeight(); ++i) {
+    for (Int i = 0; i < A.LocalHeight(); ++i)
+    {
       T val = A.GetLocal(i, j);
       expected += val * val;
     }
     expected = mpi::AllReduce(expected, g.ColComm());
     expected = Sqrt(expected);
-    if (Abs(got - expected) > 1e-5) {
+    if (Abs(got - expected) > 10 * limits::Epsilon<El::Base<T>>())
```
poulson: The accumulated error should be a function of the number of entries (with said function being logarithmic if a tree-based scheme is used, but linear with the current single-process implementation): would you mind multiplying the right-hand side by the number of entries?

Also, it would be better to use a relative bound and to diving the left-hand side by the maximum of the norm and 1 (so that it still behaves well for zero norms).

ndryden: Just want to clarify: The new check should be `Abs(got - expected) / std::max(expected, 1) > m * n * 10 * limits::Epsilon<El::Base<T>>()`?

poulson:
yes, if that isn't too much hassle. otherwise it is easy to think of cases where the check would fail.

PR #227, opened 2017-03-29
Optimizing BLAS1-like routines

### LINKED ISSUE #225: Optimizing BLAS1-like routines
I have been running numerical experiments with Axpy and I think we can significantly boost performance with some small changes. First, here are scaling studies with `float` and `Complex<double>`:
axpy_scaling_float
axpy_scaling_complexdouble
These were performed on two 12-core Intel Xeon E5-2695v2 processors. I use OpenBLAS, compile with gcc-4.9.2 -O3, construct random square test matrices, and clear the L3 cache before each trial. Note that Elemental currently calls BLAS _axpy on matrix rows or columns, depending on the matrix dimensions, so tall and square matrices correspond to "BLAS (column-wise)" and fat matrices to "BLAS (row-wise)". My takeaways:

- For `float`, column-wise BLAS calls are much faster than row-wise. This is probably since column-wise calls can coalesce memory accesses. There is no significant difference for `Complex<double>`.
- A naive implementation with nested for loops will often outperform BLAS, especially with `Complex<double>`.
- More detailed profiling is required to determining when OpenMP vectorization improves upon the naive implementation.

I'm surprised that O3 optimized Axpy so often outperforms BLAS. Once the performance of Axpy is better understood, we should be able to apply similar optimizations to similar routines (Copy, Fill, Scale, Hadamard, Scale, Shift).

timmoon10:
Resolves #225 by eliminating calls to external BLAS 1 routines. To facilitate compiler optimization, some routines iterate over a single loop rather than a double loop if the matrix data is contiguous. EntrywiseMap, IndexDependentFill, and IndexDependentMap now parallelize entry-wise (rather than column-wise) when applied to column vectors.

### DIFF
--- a/tests/blas_like/Axpy.cpp
+++ b/tests/blas_like/Axpy.cpp
@@ -1,0 +1,137 @@
+/*
+   Copyright (c) 2009-2016, Jack Poulson
+   All rights reserved.
+
+   This file is part of Elemental and is under the BSD 2-Clause License,
+   which can be found in the LICENSE file in the root directory, or at
+   http://opensource.org/licenses/BSD-2-Clause
+*/
+#include <El.hpp>
+using namespace El;
+
+template<typename T>
+void TestAxpy
+( Int m,
+  Int n,
+  Int ldimX,
+  Int ldimY,
+  Int numThreads,
+  const Grid& g,
+  bool print )
+{
+    OutputFromRoot(g.Comm(),"Testing with ",TypeName<T>());
+    PushIndent();
+
+    double runTime, gflops;
+    Timer timer;
+
+#ifdef _OPENMP
+    if( numThreads > 0 )
+    {
+        omp_set_num_threads(numThreads);
+    }
+#endif
+
+    // Generate random matrices
+    DistMatrix<T> X(g), Y(g), YOrig(g), XData(g), YData(g);
+    DistMatrix<T,STAR,STAR> alpha(g);
+    Gaussian( alpha, 1, 1 );
+    Gaussian( XData, ldimX, n );
+    View( X, XData, IR(0,m), ALL );
+    Gaussian( YData, ldimY, n );
+    View( Y, YData, IR(0,m), ALL );
+    Zeros( YOrig, m, n );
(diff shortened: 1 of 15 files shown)


tests/blas_like/Axpy.cpp
```diff
@@ -0,0 +1,137 @@
...
+        TestAxpy<DoubleDouble>( m, n, ldimX, ldimY, numThreads, g, print );
+        TestAxpy<QuadDouble>( m, n, ldimX, ldimY, numThreads, g, print );
+        TestAxpy<Complex<DoubleDouble>>( m, n, ldimX, ldimY, numThreads, g, print );
+        TestAxpy<Complex<QuadDouble>>( m, n, ldimX, ldimY, numThreads, g, print );
+#endif
+#ifdef EL_HAVE_QUAD
+        TestAxpy<Quad>( m, n, ldimX, ldimY, numThreads, g, print );
+        TestAxpy<Complex<Quad>>( m, n, ldimX, ldimY, numThreads, g, print );      
+#endif
+#ifdef EL_HAVE_QUAD
```
poulson: This line should have `EL_HAVE_MPC` rather than `EL_HAVE_QUAD`.

tests/blas_like/Axpy.cpp
```diff
@@ -0,0 +1,137 @@
...
+   Copyright (c) 2009-2016, Jack Poulson
+   All rights reserved.
+
+   This file is part of Elemental and is under the BSD 2-Clause License,
+   which can be found in the LICENSE file in the root directory, or at
+   http://opensource.org/licenses/BSD-2-Clause
+*/
+#include <El.hpp>
+using namespace El;
+
```
poulson: There seems to only be a performance test, and no correctness test. `Axpy` is pretty trivial, so perhaps this can be the exception.

timmoon10: I've added a simple correctness test where we compute Axpy on a distributed matrix, then have rank 0 compute the error at each entry. When I run the test, I get no numerical error.

PR #231, opened 2017-05-11
Optimizing BLAS 1-like routines

timmoon10:
I find that local matrix transposes take a non-trivial portion of the runtime of some variants of SUMMA GEMM (25%-50% on one process, 10% on four processes). A blocked algorithm drastically reduces time spent in these transposes.

I've also reenabled OpenMP in Axpy, Hadamard, and Scale. These were removed based on results in #225 that show that OpenMP does not make a significant improvement on 2D loops. However, tests by @ndryden on Hadamard show a 2x speedup using OpenMP on single loops.

### DIFF
--- a/include/El/blas_like/level1/Axpy.hpp
+++ b/include/El/blas_like/level1/Axpy.hpp
@@ -39,10 +39,9 @@ void Axpy( S alphaS, const Matrix<T>& X, Matrix<T>& Y )
           if( XLength != YLength )
               LogicError("Nonconformal Axpy");
         )
+        EL_PARALLEL_FOR
         for( Int i=0; i<XLength; ++i )
-        {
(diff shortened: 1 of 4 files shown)


poulson:
This is a great idea and is consistent with some changes made within the DistMatrix operator= many years ago. I can provide detailed comments after Travis finishes.

jeffhammond:
FWIW we've been benchmarking matrix transpose on Intel processors as part of the PRK project (https://github.com/ParRes/Kernels/blob/master/Cxx11/transpose-valarray.cc#L124 is the version Elemental folks are probably happiest to read) for a few years and the block size that is nearly always optimal is 32 (elements), because this ensures no TLB misses within a block.

PR #235, opened 2017-08-03
correct a path in the ElementalConfig.cmake.in file

trb413:
Both ElementalConfig.cmake and ElementalTargets.cmake are installed to ${INSTALL_CMAKE_DIR}/elemental. This is now reflected in the include() in the ElementalConfig.cmake.in file.

Also, these .cmake files will not work in the build directory. Unless someone has a more pressing need than I, I'll get to this in a later effort.

### DIFF
--- a/cmake/configure_files/ElementalConfig.cmake.in
+++ b/cmake/configure_files/ElementalConfig.cmake.in
@@ -13,6 +13,6 @@ set(Elemental_LINK_FLAGS "@EL_LINK_FLAGS@")
 set(Elemental_DEFINITIONS "@Qt5Widgets_DEFINITIONS@")
 
 # Our library dependencies (contains definitions for IMPORTED targets)
-include("@CMAKE_INSTALL_PREFIX@/CMake/elemental/ElementalTargets.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/ElementalTargets.cmake")
 
 set(Elemental_LIBRARIES El)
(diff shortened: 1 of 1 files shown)


poulson:
Thank you for this fix! Before I merge, would you mind clarifying your last statement that "these .cmake files will not work in the build directory"? Are you saying that something used to work and no longer does with this change?

trb413:
Sorry for the lack of clarity! No, the .cmake files would not have worked in the build directory anyway, with or without my change.

Some packages that build with CMake want to support users who never "install", they only "build". Thus, they might use something like [this CMake module](https://cmake.org/cmake/help/v3.0/module/CMakePackageConfigHelpers.html) to create one set of "<PackageName>Config.cmake" files in the build directory and a different set in the install directory, if the install ever happens. For an example of this, see [this code](https://github.com/mfem/mfem/blob/master/CMakeLists.txt#L426) in the MFEM library.

All I was saying was that the ElementalConfig.cmake and ElementalTargets.cmake files that get written to the Elemental build directory are not directly useable to import Elemental from its build directory -- an install must be done. If this is your intent, that's totally fine, too, and nothing else need be done. I just discovered this when I tried to pass `Elemental_DIR=/path/to/elemental/build` and the CMake of the other project failed. With this fix, passing `Elemental_DIR=/path/to/elemental/install` works fine.

trb413:
Ok so I took my own advice and looked at what I had done in MFEM. CMAKE_CURRENT_LIST_DIR is the variable that makes this stuff work in both the build and the install directory. It was a super simple change, so I did that. Should work all the time always now. 😃

poulson:
Thank you very much for clarifying! (And for taking the time to contribute the fix!)

Issue #243, opened 2017-09-11
compiler warnings


davydden:
using `Elemental` shows compiler warnings
```
/linux-centos7-x86_64/gcc-4.8.5/elemental-0.87.7-6orlczncme6fkc4675dq5ca6lrlifbwr/include/El/blas_like/level1/GetMappedDiagonal.hpp:30:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
     EL_PARALLEL_FOR
 ^
/linux-centos7-x86_64/gcc-4.8.5/elemental-0.87.7-6orlczncme6fkc4675dq5ca6lrlifbwr/include/El/blas_like/level1/GetMappedDiagonal.hpp:75:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
         EL_PARALLEL_FOR
 ^
```
(and the same for `EL_PARALLEL_FOR` elsewhere).

Plus unused variables:
```
/linux-centos7-x86_64/gcc-4.8.5/elemental-0.87.7-6orlczncme6fkc4675dq5ca6lrlifbwr/include/El/core/imports/choice.hpp:89:18: warning: unused parameter ‘os’ [-Wunused-parameter]
     virtual void HandleVersion( ostream& os=cout ) const { }
                  ^
/linux-centos7-x86_64/gcc-4.8.5/elemental-0.87.7-6orlczncme6fkc4675dq5ca6lrlifbwr/include/El/core/imports/mpi_choice.hpp:49:18: warning: unused parameter ‘os’ [-Wunused-parameter]
     virtual void HandleVersion( ostream& os=cout ) const { }
                  ^
/linux-centos7-x86_64/gcc-4.8.5/elemental-0.87.7-6orlczncme6fkc4675dq5ca6lrlifbwr/include/El/core/environment/decl.hpp:179:13: warning: unused parameter ‘os’ [-Wunused-parameter]
 inline void BuildStream( ostringstream& os ) { }
             ^
/linux-centos7-x86_64/gcc-4.8.5/elemental-0.87.7-6orlczncme6fkc4675dq5ca6lrlifbwr/include/El/core/limits.hpp:250:23: warning: unused parameter ‘alpha’ [-Wunused-parameter]
 template<> inline int Lowest<int>( const int& alpha )
                       ^
[...]

PR #244, opened 2017-09-11
Fix compiler warnings

davydden:
fixes https://github.com/elemental/Elemental/issues/243 (most of them)
~~I will double check tomorrow on Ubuntu that this indeed works (thus WIP).~~ builds on Ubuntu+gcc 5.4.0.

@poulson ping.

### DIFF
--- a/include/El/core/imports/choice.hpp
+++ b/include/El/core/imports/choice.hpp
@@ -4,8 +4,8 @@
 
    This file is a modification of Choice, a simple command-line option library.
 
-   This file is part of Elemental and is under the BSD 2-Clause License, 
-   which can be found in the LICENSE file in the root directory, or at 
+   This file is part of Elemental and is under the BSD 2-Clause License,
+   which can be found in the LICENSE file in the root directory, or at
    http://opensource.org/licenses/BSD-2-Clause
 */
 #ifndef EL_IMPORTS_CHOICE_HPP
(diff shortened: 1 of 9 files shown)


include/El/core/imports/choice.hpp
```diff
@@ -86,34 +86,35 @@ class Args
     vector<bool> usedArgs_;
     ostream& error_;
 
-    virtual void HandleVersion( ostream& os=cout ) const { }
-    virtual void HandleBuild( ostream& os=cout ) const { }
+    virtual void HandleVersion( ostream& os=cout ) const { (void)os; }
```
davydden: p.s. one could as well had those `inline void Args::...` within the header below, the current diff is just smaller.

poulson: I think defining an `EL_UNUSED` macro ala https://stackoverflow.com/a/1486931 would be more readable than the explicit voiding. Would you mind adding such a macro? This PR should be ready after that.

poulson: Also, thanks!

davydden: where do keep macros? where shall I put it?

poulson: I think in include/El/core.hpp would be most consistent.

davydden:
@poulson ping.

poulson:
There was a problem with Travis CI's backend so I just manually restarted the CI build.

PR #249, opened 2017-11-23
export El dependency on GMP, MPFR, MPC, QG, and Valgrind include dirs

evaleev:


evaleev:
Jack, thanks for a super quick turnaround!

This fix should only affect external uses of Elemental that rely on `find_package` *and* only if GMP, etc. are in a non-default prefix. So probably all Travis tests will be fine if prereqs are installed via apt-get (i.e. to /usr). Our TA Travis tests were also fine; I only had issues on my Mac where Brew installs GMP & co to /usr/local, and Apple's clang does not look there by default.

PR #250, opened 2017-11-29
Parallelizing copy and zero routines with OpenMP.

timmoon10:
I've found that we can accelerate the Copy and Zero routines by performing a manual loop with OpenMP parallelization rather than using std::memcpy or std::memset. This approach is also better for first-touch data allocation since memory will be allocated across the available NUMA domains rather than all on the first.

This change significantly improves the performance of certain Gemm configurations, although most stay roughly the same. For instance, here are the speedups I observe for the SUMMA_NN variants (matrices are square; 8 processes on 4 nodes; one 24-core Intel Xeon E5-2695 v2 per node; IB QDR interconnect):

elemental_gemm

### DIFF
--- a/include/El/blas_like/level1/Copy.hpp
+++ b/include/El/blas_like/level1/Copy.hpp
@@ -24,17 +24,29 @@ void Copy( const Matrix<T>& A, Matrix<T>& B )
     B.Resize( height, width );
     const Int ldA = A.LDim();
     const Int ldB = B.LDim();
-    const T* ABuf = A.LockedBuffer();
-          T* BBuf = B.Buffer();
+    const T* EL_RESTRICT ABuf = A.LockedBuffer();
+          T* EL_RESTRICT BBuf = B.Buffer();
 
-    // Copy all entries if memory is contiguous. Otherwise copy each
-    // column.
-    if( ldA == height && ldB == height ) {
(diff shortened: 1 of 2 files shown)


poulson:
LGTM! The factor of two speedup for float is surprising!

timmoon10:
I have some confusing timing results. I performed a scaling study on `float` arrays to compare several algorithms: "memcpy/memset" corresponds to a serial memcpy/memset call (the original algorithm for Copy/Zero), "OpenMP loop" corresponds to a simple OpenMP-parallelized loop (the current algorithm), and "OpenMP memcpy/memset" corresponds to something that looks like
```
#pragma omp parallel
{
    memcpy/memset
}
```
The L3 cache is cleared before every trial. The results are as follows:

omp_catalyst

The main takeaway is that OpenMP overhead makes it preferable to use a serial memcpy/memset when the array is small. This is not surprising, but it does bring into question why OpenMP parallelization boosts the performance of Gemm on small matrices (e.g. 16x16). My application actually slows down slightly if Copy and Zero use serial memcpy/memset on small matrices. I suspect that the OpenMP parallelized algorithm has some benefit when cache hits are possible, although I'll need to look into this more closely.

timmoon10:
If I do not clear the cache in between trials, the scaling study is as follows:

catalyst_omp_cache

PR #252, opened 2018-01-23
Buffer printing for whole matrices

mcneish1:
Hello,

Printing multiple matrices from multiple threads interleaves the elements in stdout. This buffers everything in Print that would be affected.

Please let me know if there's anything else you want me to do with this patch.


poulson:
Thanks Corey; please ignore the Travis flakes, as the tests didn't quite finish running before the 45 minute time limit kicked in.

PR #256, opened 2018-05-16
CMake, GCC: Enable libquadmath detection for GCC 5.x to 8.x.

karlrupp:
Tested with GCC 5.4 on Ubuntu 16.04.


poulson:
Thanks, Karl! Please ignore the flashing red build failure: it was due to a time out during the tests.

PR #269, opened 2018-10-25
Remove MPI runtime requirement for ParMETIS test 

ocaisa:
The test run of ParMETIS is done without the use of `MPIEXEC_EXECUTABLE` and this can lead to false negatives with some MPI implementations. This change further trivialises the test code to avoid this issue (and, given the `FIXME`, is equivalent).

### DIFF
--- a/cmake/modules/FindParMETIS.cmake
+++ b/cmake/modules/FindParMETIS.cmake
@@ -128,8 +128,8 @@ if(PARMETIS_INCLUDE_DIR AND METIS_INCLUDE_DIR AND
 int main( int argc, char* argv[] )
 {
   // FIXME: Find a simple but sensible test for ParMETIS
-  MPI_Init( &argc, &argv );
-  MPI_Finalize();
+  // MPI_Init( &argc, &argv );
+  // MPI_Finalize();
   return 0;
 }
 " PARMETIS_TEST_RUNS)


poulson:
Please ignore the Travis failure: it was just a timeout. I am not sure of the utility of removing the MPI calls (MPI is required for ParMETIS), but adding an explicit ParMETIS call would be a much-needed addition.

ocaisa:
Yes, MPI is required for ParMETIS but the existing test does not launch using `MPIEXEC_EXECUTABLE` (i.e., there is no real MPI environment). On our system with the latest Intel MPI with
```
...and run output:
Abort(1094031) on node 0 (rank 0 in comm 0): Fatal error in PMPI_Init: Other MPI error, error stack:
MPIR_Init_thread(607)......: 
MPID_Init(793).............: 
MPIDI_NM_mpi_init_hook(667): OFI addrinfo() failed (ofi_init.h:667:MPIDI_NM_mpi_init_hook:No data available)

Return value: 143
```
which is purely because the compiled code was run directly rather than via `MPIEXEC_EXECUTABLE`. I don't know CMake well enough in order to suggest something but if you want to add the ParMETIS call (which I also don't know very well, I'm just installing Elemental for someone) then you would need to fix this.

ocaisa:
Ok, no problem ;)
Feel free to close this, we will just patch the release.
Good luck with the new projects!

PR #275, opened 2019-04-10
Fix: Find Qt5 (if enabled) in ElementalConfig.cmake

JM1:
If Qt5 has been enabled during build, then e.g. target Qt5::Widgets is exported
as a install dependency via ElementalTargets.cmake. Thus ElementalConfig.cmake
must find Qt5 and import these Qt5 targets or else the user will get errors like:

CMake Error at CMakeLists.txt:... (add_executable):
  Target "..." links to target "Qt5::Widgets" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?


JM1:
`CMakeFindDependencyMacro` requires CMake 3, thus I've replaced `find_dependency` with `find_package`.

